PR #36, opened 2015-06-23
Fix py3 strings

tacaswell:
Note this is against one of your dev branches, not master.

This should not be merged to master yet as I do not think it is python2 compatible, but I needed this working on a beamline with python3 _now_ so here it is.

The source of the rouge 'b' is that the `str` or `repr` of a byte object in python 3 include the 'b' so that the string formatting `'%s%s' % (value, NULLCHAR)` -> `"b'value_str'b'\\x00"'` which is less than fully helpful.

I suspect that I have done an end-run around some of the stuff it `utils`.  Do you mind picking up a dependency on `six` (which I think would greatly simplify `utils`) or should I re-do this using existing functions (and if so, some direction as to which ones would be appreciated).

### DIFF
--- a/lib/ca.py
+++ b/lib/ca.py
@@ -1318,6 +1318,8 @@ def put(chid, value, wait=False, timeout=30, callback=None,
 
     else:
         if ftype == dbr.CHAR and is_string_or_bytes(value):
+            if isinstance(value, bytes):
+                value = value.decode('ascii', 'replace')
             value = [ord(i) for i in ("%s%s" % (value, NULLCHAR))]
         try:
             ndata, nuser = len(data), len(value)
--- a/lib/ca.py
+++ b/lib/ca.py
@@ -1300,12 +1300,14 @@ def put(chid, value, wait=False, timeout=30, callback=None,
     elif nativecount == 1:
         if ftype == dbr.CHAR:
             if is_string_or_bytes(value):
-                value = [ord(i) for i in ("%s%s" % (value, NULLCHAR))]
+                if isinstance(value, bytes):
+                    value = value.decode('ascii', 'replace')
+                value = [ord(i) for i in value] + [0, ]
             else:
                 data[0] = value
         else:
             # allow strings (even bits/hex) to be put to integer types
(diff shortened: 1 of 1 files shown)


newville:
@tacaswell  Thanks.  I agree with the assessment, and I'll try this out here too soon.  I'm a little cautious about adding a dependency, but requiring `six` is probably OK.

tacaswell:
If you are very wary of picking up a dependency bundling a private version of six is also an option (see https://github.com/matplotlib/matplotlib/pull/4501 ).  I am not a fan of this approach but it is better than writing a pyepics-only version of the same functionality.

tacaswell:
@newville Just to let you know, we are doing something sort of bad-form, but it is the cleanest way to get what we need.

I have forked this to the NSLS-II org and pushed the tag `'v3.2.4.post19+g763d30'` https://github.com/NSLS-II/pyepics/releases/tag/v3.2.4.post19%2Bg763d30 so that we have a fixed target to build our internal binaries against.  The naming on this pep440 compliant and will sort after the official `3.2.4` and before a future official `3.2.5`.

I do not intend this tag to have any use other than our internal use.

We could have had our build products against a SHA, but this is more readable.

newville:
@tacaswell That's fine with me.    And, sorry, this PR sort of fell off my radar screen.   I'll merge right away.  

Would it be better to add a tag to the main pyepics repo?  If you'd like, I'd be happy to add you or anyone else at NSLS-II as project developers / owners.

tacaswell:
This breaks on Python 2

newville:
@tacaswell  Oh, wait, what is broken?   The unittests work for me with Python 2.7, thought I haven't tested everything on all platforms.

newville:
@tacaswell   .... the unit tests pass for me with Python2 on the fix_py3_strings branch too....  Maybe I'm missing some tests.

tacaswell:
interesting, learned something I should have known.  `unicode` is not in py3, but it turns out `bytes` _is_ in py2.7, good to know. :sheep: 

Sorry for the noise.

I see you already have @klauer, can you add me and @dchabot as well?

PR #43, opened 2015-10-01
units attribute needs BYTE2STR()

### LINKED ISSUE #42: `pv.units` returns a byte string on py3.x
via https://github.com/NSLS-II/ophyd/issues/135#issuecomment-144537856

> One thing @dchabot and I found yesterday was that epics.PV().units returns a byte string in 3.x. Need to get this fixed upstream in pyepics.

Moved here so it does not get lost.

newville:
fixes #42, I believe.


newville:
@dchabot OK, done!  Thanks for finding this.

PR #45, opened 2015-10-09
Search for libca.so in EPICS_HOST_ARCH also.

NickeZ:
We don't use the regular "linux-" prefix, but instead different depending on operating system. E.g. SL6, centos7 and so on. Therefore we depend on that you actually use EPICS_HOST_ARCH.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -159,9 +159,10 @@ def find_libca():
             libname = 'libca.dylib'
 
         epics_base = os.environ.get('EPICS_BASE', '.')
+        epics_host_arch = os.environ.get('EPICS_HOST_ARCH')
         host_arch = os.uname()[0]
         if host_arch in known_hosts:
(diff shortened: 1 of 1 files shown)


tacaswell:
attn @dchabot

newville:
@NickeZ  Hmm, I take it that is preferable to setting 'PYEPICS_LIBCA' or LD_LIBRARY_PATH?   I could be persuaded of that.  But it sort of seems like EPICS_HOST_ARCH is meant for building cross-platform, not for describing the OS flavor.   Is EPICS_HOST_ARCH really the best approach?

NickeZ:
Different linux distros use different libc for example, and this is one way to have them all in the same shared file system. I know of more labs who also use it like this.

If you dont accept this PR i would kindly ask you to change the documentation.

newville:
Thanks!

PR #46, opened 2015-11-25
fix typo in argument default

birkenfeld:


newville:
Good catch -- thanks!

PR #47, opened 2015-12-09
Make get_pv respect the form argument when creating new PVs

RobbieClarken:
The `get_pv` function accepts the `form` parameter but was not using it when creating new PVs. This meant `Device.add_pv` was unable to create PVs with `form='ctrl'` for example. This pull request fixes this by passing in the `form` parameter when instantiating a new PV.

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -43,7 +43,7 @@ def get_pv(pvname, form='time',  connect=False,
     start_time = time.time()
     # not cached -- create pv (automaticall saved to cache)
     if thispv is None:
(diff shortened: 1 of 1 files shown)


newville:
@RobbieClarken  Good catch!  Thanks!

RobbieClarken:
Thanks for merging! When you have time could you you please upload 3.2.5rc3 to pypi so I can include it in my setup.py dependencies.

newville:
@RobbieClarken  Thanks for the reminder.  Uploaded to PyPI.

PR #48, opened 2016-01-19
FIX: fix two bugs handling empty strings

tacaswell:
- ascii_string correctly handles `''` and does not need to be
 short-cirucited

- if the length of the input is 0, then the length is re-set to the
maximum length of the array.  Later, there is a check that if ftype ==
dbr.CHAR then add 1 to count.  This results in a error from trying to
put a too-long array to the PV.

There are still bugs in handling the edge cases of very long
strings (the null char might not get written).

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -1290,11 +1290,10 @@ def put(chid, value, wait=False, timeout=30, callback=None,
      value put() to array PV must be an array or sequence''')
     if ftype == dbr.CHAR and nativecount > 1 and is_string_or_bytes(value):
         count += 1
+        count = min(count, nativecount)
 
     # if needed (python3, especially) convert to basic string/bytes form
     if is_string(value):
-        if value == '':
-            value = NULLCHAR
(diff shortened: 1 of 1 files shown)


tacaswell:
I think this needs to go into the next release if that is not too late and we will be deploying this code at BNL later today.

attn @klauer

newville:
@tacaswell @klauer  OK, I will merge and run a full battery of tests to get 3.2.5 released soon.  I'm sort of swamped with lots of other stuff. My time-frame would be end of next week -- is that too slow for you all?  

The use of string data is sort of complicated in CA.   The epics datatype really maps to bytes, and does not really require a terminating NULL character.    In addition,  there are single character types (PROC fields, for example) that need to be handled, so some of the code is a bit more complicated and fragile than I'd like.  FWIW,  using Char Waveform as Long String sort of works,  but they aren't really the same things. NULL-termination as for strings is the root of many problems.

Anyway, I think this PR is fine.  I'll merge soon.

PR #49, opened 2016-01-25
FIX: zero length subarray

klauer:
In testing at NSLS-II, we found that uninitialized (undefined/zero-length) subarray records were returning a ctypes.array of length 0 in some cases. This PR's changes in `epics/ca.py`'s `unpack` functions should be sufficient to fix this. Included is also a unit test to check this result.

@newville, I made an additional change here that may have implications elsewhere, which I think makes sense. Feel free to correct me if I'm missing something, though.

A simple script using an uninitialized subArray PV with 64 elements in the current pyepics master:
```python
element count according to ca 64
get(use_monitor=True, count=None)=<epics.dbr.c_double_Array_0 object at 0x7fc96ff51680>
get(use_monitor=False, count=None)=array([  2.50320809e-308,   0.00000000e+000,   0.00000000e+000,
         0.00000000e+000,   0.00000000e+000,   0.00000000e+000,
         0.00000000e+000,   0.00000000e+000,   0.00000000e+000,
         0.00000000e+000]) (first 10 elements)
```
Both of these are incorrect - the first because it returns a ctypes array instead of a numpy array. The second is incorrect (in my opinion), because it returns the full array despite libca knowing that it is of 0-length.

Note that when `use_monitor=True`, the element count is set in the monitor callback - which allows `get(use_monitor=True)` to return the correct length array.

The same script output, using this PR:
```python
element count according to ca 64
get(use_monitor=True, count=None)=array([], dtype=float64)
get(use_monitor=False, count=None)=array([], dtype=float64) (first 10 elements)
```

The difference for `use_monitor=False` comes in in `epics.ca.get()` - if the user doesn't specify `count`, I modified it to use the length reported by EPICS rather than using the maximum element length of the channel (i.e., `ca.element_count`). (ca.py line 1135)

The only argument against this I can see is that EPICS command line tools behave more like the current master (however unintuitive it may be):
```sh
$ caget 'Py:ZeroLenSubArr1'
Py:ZeroLenSubArr1 64 2.50321e-308 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
$ caget -c 'Py:ZeroLenSubArr1'
Py:ZeroLenSubArr1
$ camonitor 'Py:ZeroLenSubArr1'
Py:ZeroLenSubArr1              <undefined> UDF INVALID
```


newville:
@klauer   Thanks!! This looks good to me, and I'll merge.  I don't know that Epics CA actually specifies a "correct" behavior here, and what the command line tools is a reasonable guide, but so is doing the most Pythonic thing.   

Returning the ctype was definitely because the code was just falling through and giving up on not being able to figure out what to do.   Anyway, I agree that your choices are the right ones.

PR #50, opened 2016-01-25
Fix: Py3k waveform of strings put()

klauer:
* Fixes waveform of strings `put()` in Python 3, added a unit test for it. It was just a missing a call to `ascii_string`. 
* Adds a context manager to pause simulator updating for the tests. I found that if one test failed it might affect the others, as simulator updating was never resumed
* All tests are passing now in 2.7/3.4 (minus the 'real motor' one which I don't have an IOC for - I added a skiptest for this in my other PR #49).

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -98,7 +98,9 @@ class ChannelAccessException(Exception):
     """Channel Access Exception: General Errors"""
     def __init__(self, *args):
         Exception.__init__(self, *args)
-        sys.excepthook(*sys.exc_info())
+        type_, value, traceback = sys.exc_info()
+        if type_ is not None:
+            sys.excepthook(type_, value, traceback)
 
 class CASeverityException(Exception):
     """Channel Access Severity Check Exception:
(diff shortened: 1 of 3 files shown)


newville:
@klauer  we should probably remove "real motor" (I often set it up to really run an actual motor, and then can be very sad when I forget to turn this off!!).   I believe this could be simulated with an Epics Record that processes slowly, emulating something not being done immediately.

newville:
@klauer   ... are these conflicts resolvable?   What's the relation with testioc?

klauer:
Rebased with master and force-pushed (git couldn't figure out how to merge a single additional blank line...)

Regarding the real motor test - yes, I think it can be done better. At NSLS-II we're using the soft motor support built into the motor record for this purpose. Perhaps we can modify the test IOC to do this in the future.

This changeset makes it such that calling `.encode()` on all waveforms of strings is no longer necessary. So, if this gets merged, the simulator PR should be merged, too.

newville:
@klauer, @tacaswell    I'm merging both of these now.  Thanks so much!

I think it's time to tag and release 3.2.5.  Any objections?

klauer:
No objections from me.

PR #53, opened 2016-02-04
Fix scalar len and zero-length strings

klauer:
I ran into further issues with zero-length strings and some scalar-value related oddities. Here are two small fixes: both should be minor enough as to not break anything else. Added 2 tests for these. Cousin PR is pyepics/testioc#3.

Example PV:
```sh
$ caget -Sc 'XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV'
XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV 0
```

Case 1 that was failing (non-automonitored waveform char as string):
```python
In [1]: epics.PV('XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV', auto_monitor=False).get(as_string=True)

pyepics/epics/pv.py in get(self, count, as_string, as_numpy, timeout, with_ctrlvars, use_monitor)
    290         val = self._args['value']
    291         if as_string:
--> 292             return self._set_charval(val)
    293         if self.count <= 1 or val is None:
    294             return val

pyepics/epics/pv.py in _set_charval(self, val, call_ca)
    360             elif self.count==1: # handles single character in waveform
    361                 val = [val]
--> 362             val = list(val)
    363             if 0 in val:
    364                 firstnull  = val.index(0)

TypeError: 'int' object is not iterable
```

Case 2:
```python
In [2]: epics.PV('XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV', auto_monitor=False).get()

pyepics/epics/pv.py in get(self, count, as_string, as_numpy, timeout, with_ctrlvars, use_monitor)
    295
    296         if count is None:
--> 297             count = len(val)
    298         if (as_numpy and ca.HAS_NUMPY and
    299             not isinstance(val, ca.numpy.ndarray)):

TypeError: object of type 'int' has no len()
```

With this PR:
```python
In [1]: epics.PV('XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV', auto_monitor=False).get()
Out[1]: 0

In [2]: epics.PV('XF:31IDA-BI{Cam:Tbl}cam1:StringFromServer_RBV', auto_monitor=False).get(as_string=True)
Out[2]: ''

```

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -1031,13 +1031,15 @@ def _unpack(chid, data, count=None, ftype=None, as_numpy=True):
             else:
                 out = numpy.ctypeslib.as_array(memcopy(data))
         else:
-            out = memcopy(data)
+            out = list(data)
         return out
 
-    def unpack(data, count, ntype, use_numpy):
+    def unpack(data, count, ntype, use_numpy, elem_count):
         "simple, native data type"
         if data is None:
             return None
+        elif ntype == dbr.CHAR and elem_count > 1:
+            return array_cast(data, count, ntype, use_numpy)
         elif count == 1 and ntype != dbr.STRING:
             return data[0]
         elif ntype == dbr.STRING:
@@ -1062,8 +1064,11 @@ def _unpack(chid, data, count=None, ftype=None, as_numpy=True):
     if ftype is None:
         ftype = dbr.INT
     ntype = native_type(ftype)
-    use_numpy = (HAS_NUMPY and as_numpy and ntype != dbr.STRING and count != 1)
-    return unpack(data, count, ntype, use_numpy)
+    elem_count = element_count(chid)
+    numpy_count = (count != 1 and elem_count > 1)
+    use_numpy = (HAS_NUMPY and as_numpy and ntype != dbr.STRING and
+                 numpy_count)
+    return unpack(data, count, ntype, use_numpy, elem_count)
 
 @withConnectedCHID
 def get(chid, ftype=None, count=None, wait=True, timeout=None,
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -294,23 +294,16 @@ class PV(object):
             return val
 
         if count is None:
-            try:
-                count = len(val)
-            except TypeError:
-                # not an array-type, just return the value itself
-                return val
+            count = len(val)
 
-        if (as_numpy and ca.HAS_NUMPY and
-            not isinstance(val, ca.numpy.ndarray)):
-            if count == 1:
-                val = [val]
-            val = ca.numpy.array(val)
-        elif (as_numpy and ca.HAS_NUMPY and count == 1 and
-            not isinstance(val, ca.numpy.ndarray)):
-            val = ca.numpy.array([val])
-        elif (not as_numpy and ca.HAS_NUMPY and
-              isinstance(val, ca.numpy.ndarray)):
-            val = list(val)
+        if ca.HAS_NUMPY:
+            is_array = isinstance(val, ca.numpy.ndarray)
+            if as_numpy and not is_array:
+                if not isinstance(val, list):
+                    val = [val]
+                val = ca.numpy.array(val)
+            elif not as_numpy and is_array:
+                val = val.tolist()
         # allow asking for less data than actually exists in the cached value
         if count < len(val):
             val = val[:count]
@@ -360,15 +353,13 @@ class PV(object):
             return val
         # char waveform as string
         if ntype == dbr.CHAR and self.count < ca.AUTOMONITOR_MAXLENGTH:
-            if isinstance(val, ca.numpy.ndarray):
+            if ca.HAS_NUMPY and isinstance(val, ca.numpy.ndarray):
                 val = val.tolist()
-            elif self.count==1 or isinstance(val, int):
-                # single character in waveform
+            elif not isinstance(val, list):
                 val = [val]
-            else:
-                val = list(val)
+
             if 0 in val:
-                firstnull  = val.index(0)
+                firstnull = val.index(0)
             else:
                 firstnull = len(val)
             try:
@@ -379,12 +370,9 @@ class PV(object):
             return cval
 
         cval = repr(val)
+
         if self.count > 1:
-            try:
-                array_size = len(val)
-            except TypeError:
-                array_size = None
-            cval = '<array size=%s, type=%s>' % (array_size,
+            cval = '<array size=%s, type=%s>' % (len(val),
                                                  dbr.Name(ftype).lower())
         elif ntype in (dbr.FLOAT, dbr.DOUBLE):
             if call_ca and self._args['precision'] is None:
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -294,7 +294,12 @@ class PV(object):
             return val
 
         if count is None:
-            count = len(val)
+            try:
+                count = len(val)
+            except TypeError:
+                # not an array-type, just return the value itself
+                return val
+
         if (as_numpy and ca.HAS_NUMPY and
             not isinstance(val, ca.numpy.ndarray)):
             if count == 1:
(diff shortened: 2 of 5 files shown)


newville:
@klauer Sorry if this seems overly picky, but I'm not sure I understand the case for how one gets to a point where

```   elif self.count==1 or isinstance(val, int): ``` 

is needed in PV._set_charval()

I see this (Python2.7.10):

```python

>>> import epics
>>> epics.__version__
'3.2.5rc1-19-g342c09e'
>>> p = epics.PV('Py:char128', auto_monitor=False)
>>> p.put([0])
1
>>> p
<PV 'Py:char128', count=1, type=time_char, access=read/write>
>>> p.value
0
>>> p.get(as_string=True)
''

```

That is, I don't see an obvious problem.  I also don't get the later try/except clause around `array_size = len(val)`.  Am I missing something?

I can certainly believe that more test cases are needed for using character waveforms as strings.

klauer:
You absolutely should be picky - I may not have fixed things in the right places.

Try cherry-picking the tests from this PR and run them with the current master. I believe they will fail. If not, I'll close the PR and call it a day...

newville:
@klauer  my point is that I'm not able to reproduce the alleged problem.    I have a char array that has a value of 0, and I get the string '' returned.

I don't understand how does one get into a situation with a CHAR array where the value is an integer, and count is not 1?  And in that case, what is the right solution?

klauer:
Well, I put together a test case which you sidestepped by running your own interactive session. I can't say as to why your results are different from mine nor why this issue appears in the first place.

In any case, repeating your test by distilling the unittest I submitted and replacing the PV name with one already in the simulator, I see the same result on 2.7/3.4 with current pyepics master c5c09b9b395e0118140683b1fec1ccd00f404c0f:

`test.py`:
```python
from epics import PV

# pvname = 'Py:waveform_char256'  # <-- this has the same result
pvname = 'Py:char128'

zerostr = PV(pvname, auto_monitor=False)
# zerostr.wait_for_connection()  # no effect

zerostr.put([0])
print(zerostr)

assert zerostr.get(as_string=True) == ''
assert zerostr.get(as_string=False) == 0
```

I see this:
```sh
$ python test.py
<PV 'Py:char128', count=128/128, type=time_char, access=read/write>
  File "test.py", line 15, in <module>
    assert zerostr.get(as_string=True) == ''
  File "epics/pv.py", line 292, in get
    return self._set_charval(val)
  File "epics/pv.py", line 362, in _set_charval
    val = list(val)
TypeError: 'int' object is not iterable
```

A difference appears to show up in `count` from your output (mine shows 128/128, yours shows 1/1). Would you please try running this example script?

If possible, it would be best to fix this at the source than bandaging up where it's failing - I'm just not sure where that is.

newville:
@klauer  Sorry for delay, kind of a crazy morning.  But, yes, with master this also fails for me!  Good, we have a clear problem, though I didn't investigate carefully. Don't know why the other approach didn't fail.

klauer:
@newville Not a problem - good that we're on the same page now. As far as how to fix it (properly?) that's another story...

newville:
@klauer  git bisect suggests 3f56c02d79fc6ebb2444fbf0228bd44911ea4121  to be the culprit.

That is, before that change, the 2nd assertion fails because the non-string value is the null array of 0s, not a single integer with value 0.   So, maybe we just have to handle array values that are length 1 and 0 slightly differently?   I'm not sure, and could probably be persuaded otherwise, but I think I would prefer the value in this case to be [0] (with as_string returning ''), instead of 0.

klauer:
It doesn't surprise me that I broke it initially! I think you're right, `[0]` and `''` rather than `0` and `''`.

How do you differentiate the meaning of `count=1` to be an array or a scalar value? Would something like this work?
```python
if count == 1:
    if element_count(chid) > 1:
       return array_of_len_1
    return scalar
```

newville:
@klauer  Yes, I think  that `ca._unpack.unpack()` should be changed to look for `count==1 and element_count > 1`.    I think that this might work (for CHAR waveforms that happen to have 1 element AND for CHAR values with count=1 like PROC fields), though I haven't fully tested it:

ca.py, line 1038

```python
    def unpack(data, count, ntype, use_numpy, ncount):
        "simple, native data type"
        if data is None:
            return None
        elif count == 1 and ntype == dbr.CHAR and ncount > 1:
            return array_cast(data, count, ntype, use_numpy)
        elif count == 1 and ntype != dbr.STRING:
            return data[0]
        elif ntype == dbr.STRING:
            return scan_string(data, count)
        elif count > 1:
            return array_cast(data, count, ntype, use_numpy)
        return data

    # Grab the native-data-type data
    try:
        data = data[1]
    except (TypeError, IndexError):
        return None

    if count == 0 or count is None:
        count = len(data)
    else:
        count = min(len(data), count)

    if ftype is None and chid is not None:
        ftype = field_type(chid)
    if ftype is None:
        ftype = dbr.INT
    ntype = native_type(ftype)
    use_numpy = (HAS_NUMPY and as_numpy and ntype != dbr.STRING and count > 1)
    return unpack(data, count, ntype, use_numpy, element_count(chid))
``` 

Note the addition of the element count to unpack call.   Suggestions welcome!

klauer:
I updated the branch based on your suggestion - trying to clean up the logic in places. It largely removes the bad fixes I initially made.

I made what could potentially be one controversial change - there were cases where raw ctypes arrays were being returned. I thought this was undesirable, but there was a test for this in `ca_unittest.py` which make it seem intentional. 

The rest should probably be looked through carefully to ensure I didn't miss anything with the logic. It gets pretty confusing... Tests passing on 2.7/3.4.

epics/ca.py
```diff
@@ -1031,13 +1031,15 @@ def array_cast(data, count, ntype, use_numpy):
...
-            out = memcopy(data)
+            out = list(data)
         return out
 
-    def unpack(data, count, ntype, use_numpy):
+    def unpack(data, count, ntype, use_numpy, elem_count):
         "simple, native data type"
         if data is None:
             return None
+        elif ntype == dbr.CHAR and elem_count > 1:
```
klauer: Proposed solution was `if count == 1 and ntype == dbr.CHAR and ncount > 1:`, but I believe we always want to get an array back if it's a char waveform (so a value of '\0\0' with a count of 2 would give back an array too)

newville: Is count == 1 in that case? 

If count > 1, it's an array for sure. If count==1 it *might* be an array.

epics/ca.py
```diff
@@ -1031,13 +1031,15 @@ def array_cast(data, count, ntype, use_numpy):
             else:
                 out = numpy.ctypeslib.as_array(memcopy(data))
         else:
-            out = memcopy(data)
+            out = list(data)
```
newville: Why replace memcopy() (which should be copy.copy()) with list() here? 

Does this have anything to do with zero-length strings?  What else with this change break?

epics/ca.py
```diff
@@ -1062,8 +1064,11 @@ def unpack(data, count, ntype, use_numpy):
     if ftype is None:
         ftype = dbr.INT
     ntype = native_type(ftype)
-    use_numpy = (HAS_NUMPY and as_numpy and ntype != dbr.STRING and count != 1)
-    return unpack(data, count, ntype, use_numpy)
+    elem_count = element_count(chid)
+    numpy_count = (count != 1 and elem_count > 1)
```
newville: is there a case where count != 1  and not(count >1) and elem_count > 1?

newville:
@klauer this seems to have a lot of changes that are not obvious relevant to the issue at hand and that seem pretty scary. There are many subtle bits in ca.py.  

Small, clear PRs that fix a single problem are much preferred.

epics/pv.py
```diff
@@ -355,13 +353,13 @@ def _set_charval(self, val, call_ca=True):
             return val
         # char waveform as string
         if ntype == dbr.CHAR and self.count < ca.AUTOMONITOR_MAXLENGTH:
-            if isinstance(val, ca.numpy.ndarray):
+            if ca.HAS_NUMPY and isinstance(val, ca.numpy.ndarray):
```
newville: what if one doesn't have NUMPY installed?   Doesn't this test now FAIL HARD?
What does this have to do with the issue?

epics/pv.py
```diff
@@ -295,17 +295,15 @@ def get(self, count=None, as_string=False, as_numpy=True,
...
-            val = list(val)
+
+        if ca.HAS_NUMPY:
+            is_array = isinstance(val, ca.numpy.ndarray)
+            if as_numpy and not is_array:
+                if not isinstance(val, list):
+                    val = [val]
+                val = ca.numpy.array(val)
+            elif not as_numpy and is_array:
+                val = val.tolist()
```
newville: Is this change meant to be a cleanup for clarity?  Because the behavior is actually different.

klauer:
I'll just close this PR - it's getting out of hand.

klauer:
I'll see if I can come up with a cleaner fix with less potential implications some other day.

PR #55, opened 2016-02-18
Fix "emptyish" char waveforms

klauer:
A second and much more focused attempt at fixing #53. Does not require simulator or test IOC changes.

I'll be addressing comments from the last PR inline with the code shortly.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -1035,10 +1035,12 @@ def _unpack(chid, data, count=None, ftype=None, as_numpy=True):
             out = memcopy(data)
         return out
 
-    def unpack(data, count, ntype, use_numpy):
+    def unpack(data, count, ntype, use_numpy, elem_count):
         "simple, native data type"
         if data is None:
             return None
+        elif ntype == dbr.CHAR and elem_count > 1:
+            return array_cast(data, count, ntype, use_numpy)
         elif count == 1 and ntype != dbr.STRING:
             return data[0]
         elif ntype == dbr.STRING:
@@ -1063,8 +1065,9 @@ def _unpack(chid, data, count=None, ftype=None, as_numpy=True):
     if ftype is None:
         ftype = dbr.INT
     ntype = native_type(ftype)
+    elem_count = element_count(chid)
     use_numpy = (HAS_NUMPY and as_numpy and ntype != dbr.STRING and count != 1)
-    return unpack(data, count, ntype, use_numpy)
+    return unpack(data, count, ntype, use_numpy, elem_count)
 
 @withConnectedCHID
 def get(chid, ftype=None, count=None, wait=True, timeout=None,
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -28,7 +28,7 @@ def get_pv(pvname, form='time',  connect=False,
     context   PV threading context (default None)
     timeout   connection timeout, in seconds (default 5.0)
     """
-    
+
     if form not in ('native', 'time', 'ctrl'):
         form = 'native'
 
@@ -39,7 +39,7 @@ def get_pv(pvname, form='time',  connect=False,
             context = ca.current_context()
         if (pvname, form, context) in _PVcache_:
             thispv = _PVcache_[(pvname, form, context)]
-    
+
     start_time = time.time()
     # not cached -- create pv (automaticall saved to cache)
     if thispv is None:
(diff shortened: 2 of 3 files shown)


epics/ca.py
```diff
@@ -1035,10 +1035,12 @@ def array_cast(data, count, ntype, use_numpy):
             out = memcopy(data)
         return out
 
-    def unpack(data, count, ntype, use_numpy):
+    def unpack(data, count, ntype, use_numpy, elem_count):
         "simple, native data type"
         if data is None:
             return None
+        elif ntype == dbr.CHAR and elem_count > 1:
```
klauer: Original comment was:
> Is count == 1 in that case?
> If count > 1, it's an array for sure. If count==1 it might be an array.

As a reminder: `elem_count` is the total element count of the channel, not necessarily the size of the current value held within. `count` is the client-requested number of elements. When the client requests 0 (or `None`) elements, `count` gets automatically set to the size of the current value (which is `<= elem_count`).

With that, I believe that the `elem_count` as reported by EPICS is the key here, not the `count`. If the EPICS layer says that this as a char array and not a single char value (i.e., `elem_count > 1`), we should always get a char array back.

epics/pv.py
```diff
@@ -356,10 +356,22 @@ def _set_charval(self, val, call_ca=True):
...
             if isinstance(val, ca.numpy.ndarray):
+                # a numpy array
                 val = val.tolist()
-            elif self.count==1: # handles single character in waveform
-                val = [val]
-            val = list(val)
+
+                if not isinstance(val, list):
+                    # a scalar value from numpy, tolist() turns it into a
+                    # native python integer
```
klauer: As the comment mentions, we can sometimes get a single scalar value back from `tolist()`:
```python
In [1]: import numpy as np

In [2]: scalar_value = np.float64(1)

In [3]: scalar_value.tolist()
Out[3]: 1.0

In [4]: type(scalar_value.tolist())
Out[4]: float
```

This breaks the ability to search for a null terminator, hence the check for the list type here.

epics/pv.py
```diff
@@ -356,10 +356,22 @@ def _set_charval(self, val, call_ca=True):
...
+                    val = [val.tolist()]
+            else:
+                try:
+                    # otherwise, try forcing it into a list. this will fail for
+                    # scalar types
+                    val = list(val)
```
klauer: `list()` here should work in the case of a ctypes array:
```python
In [2]: byte_array = (ctypes.c_ubyte * 5)()

In [3]: list(byte_array)
Out[3]: [0, 0, 0, 0, 0]
```

but fail in the case of a scalar value:
```python
In [6]: list(ctypes.c_ubyte())   # raises TypeError
```

epics/pv.py
```diff
@@ -356,10 +356,22 @@ def _set_charval(self, val, call_ca=True):
         # char waveform as string
         if ntype == dbr.CHAR and self.count < ca.AUTOMONITOR_MAXLENGTH:
             if isinstance(val, ca.numpy.ndarray):
```
klauer: I attempted to fix this in the other PR - but am purposefully leaving it out in this one to limit the scope. 

This will fail miserably without numpy installed as there is no check for it. I think it would be a good idea to either consider making numpy a requirement for pyepics, or always run tests in two separate environments - with and without numpy.

klauer:
Sorry, this still is somewhat incomplete.

There's at least one remaining, related bug that needs fixing. I'll try to update it shortly.

newville:
@klauer Thanks for coming back to this!  I think I agree with all the points you make.  

For numpy being a requirement, it might be acceptable to get less-than-ideal results (like, a ctypes values) when dealing with waveforms (especially char waveforms) if numpy is not available.   For sure, any workstation or server really using pyepics is definitely going to want numpy installed anyway.   I'm less sure about raspberry pis and other small devices.  I *think* installing numpy is not a burden, but the use of numpy in pyepics can be pretty marginal if you're interfacing to a motor or running a temperature controller.  Basically, I'm not sure.

klauer:
@newville not a problem. I hope this PR better addresses the issue. I have yet to see any issues where people complain about numpy-less environments working, so chances are people are installing it. It's certainly not important for us at the moment, though.

Anyway, I've added one more commit, which I did my best to keep to a minimum of changes. It fixes up the edge cases (that aren't quite clear to me why they happen...) where unpack() is returning a scalar value yet `epics.PV` is expecting an array value. This leads to both `PV._set_charval` and `PV.get` failing. If you'd like to see those issues for yourself to propose a less band-aided fix, take the new test `pv_unittest.test_subarray_1elem` and run it against 497f710. 

The PR as of dea8b1c is passing both ca/pv_unittest and a local NSLS-II test that was failing.

newville:
@klauer  OK, ready for merge now or should we wait a day?

klauer:
@newville I'd say if there are no objections by tomorrow, let it go through. In the meantime, I'm going to try to have @tacaswell to poke around on our end to make sure I didn't miss something.

newville:
@klauer OK, merged.... Thanks!

Issue #56, opened 2016-03-01
Inconsistencies with status and severity


NomAnor:
Hi,
I'm monitoring some records and I was wondering why I got a status value of 1 in the callback while a caget on the command line gave me no alarm status. When creating a PV, `form='time'` and `form='ctrl'` give different values (with different meanings) in the pv.status attribute.

After digging around in the code I found the problem. In the file ca.py in the function _onMonitorEvent the status keyword is assigned the value of the status argument (Line 465). According to the EPICS Channel Access manual (http://www.aps.anl.gov/epics/base/R3-14/12-docs/CAref.html#ca_add_event) this is the status of the call (1 for sucess in caerr.h). The pyepics manual (http://pyepics.github.io/pyepics/pv.html#attributes) states that 1 is the ok value but this is not the alarm status of the PV.

Later in the function if someone uses `form='time'`, the status keyword is overwritten with the record alarm status value (kwds['status'] = tmpv.status). This is what I would expect in all circumstances.

I propose to change line 465 to
`kwds = {'ftype':args.type, 'count':args.count, 'chid':args.chid, 'pvname': pvname}`
and line 475 to:
`for attr in dbr.ctrl_limits + ('precision', 'units', 'status', 'severity'):`
And change the manual accordingly.

The pv.status and pv.severity attributes (and the keywords in the callbacks) should always be the alarm values.

newville:
@NomAnor  Good point.  I have to admit rarely paying attention to status and severity fields.  What *is* `args.status` holding if not, you know, "status"?  It's described as
"ECA_XXX status of the requested op from the server".   
I'm not sure I really understand why (when) that is different from "status"?

Anyway, I'd be OK with your proposed changes.   Would you like to make a PR for the changes to code and doc?

klauer:
The status returned is the status of the call - `ECA_NORMAL` and such, but the status of the PV is alarm status/severity. Alarm information comes in all but the primitive/native type requests (so not `DBR_FLOAT` but `DBR_STS_FLOAT`, `DBR_TIME_FLOAT`, `DBR_CTRL_FLOAT`.)

The suggested changes sound good.

newville:
@NomAnor @klauer  OK, that `args.status` doesn't sound highly useful to me (like, the monitor got called.  Is it worth Should we preserve the value of `args.status` someplace, perhaps 'monitor_status'?    We're not saving this for all data type variants, but perhaps we should?   I'm fine either way.

[...]

PR #57, opened 2016-03-02
FIX: Always return the PV alarm status in the pv.status attribute

NomAnor:
Fix for #56


newville:
@NomAnor Thanks!  I'll merge this.  We can consider further changes as in the discussion at #56 if needed.

PR #58, opened 2016-03-03
API: If precision is not given, return None, not 0.

danielballan:
@newville Would you be open to this change? Downstream, we'd like to display 3 decimal places by default when precision is not defined by EPICS. But pyepics currently doesn't provide enough information for us to know whether precision is undefined or actually 0.

cc @dchabot You and I [had the same thought](https://github.com/NSLS-II/ophyd/pull/284#issuecomment-191796313).

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -1488,8 +1488,8 @@ def get_precision(chid):
     """return the precision of a Channel.  For Channels with
     native type other than FLOAT or DOUBLE, this will be 0"""
     if field_type(chid) in (dbr.FLOAT, dbr.DOUBLE):
-        return get_ctrlvars(chid).get('precision', 0)
-    return 0
+        return get_ctrlvars(chid).get('precision', None)
+    return None
 
 def get_enum_strings(chid):
     """return list of names for ENUM states of a Channel.  Returns


klauer:
Although I do agree with the change, there is always a chance that someone is relying on this returning `0`...

danielballan:
Right. I almost added: if you sense this is going to break a bunch of downstream code, we can find another way.

newville:
@danielballan  @klauer  @dchabot    I could be persuaded by this, but I'm not sure.

Does it happen often that get_precision() returns 0 and/or that get_ctrlvars(chid) didn't get a value for precision for floats?   Maybe *that* should be solved?   

Returning `None` to signal "I don't know the precision" puts the formatting burden on every end user and they would have to test for `None` and make a choice.   As it is, they get a usable default value.   In fact, you are sort of suggesting that the unknown value should be 3, which has the benefit of being a valid value. 

Another option would be to not give a default value at all and allow an exception to be raised that the user would be forced to deal with.   That might be clearer than `None`, actually, if slightly more obnoxious.

Anyway, I don't really have a strong opinion.  I almost always use 'pv.precision', which can be `None`, but if so, I tend to assume that "unknown" should be 0.  3 would be OK too.

danielballan:
> Does it happen often that get_precision() returns 0 and/or that get_ctrlvars(chid) didn't get a value for precision for floats? Maybe that should be solved?

100% agree that EPICS should be providing a precision -- papering over it in Python should be never be Plan A.

Nonetheless, I'd like utilities downstream of pyepics to be able to choose their own default, or choose to raise. In that regard, `None` is more flexible than a hard-coded default, whether it be 0 or 3.

newville:
@danielballan Yeah, I guess this (returning `None` to signal that `ca.get_precision()` failed) is OK.   I think I never actually use this function and rely solely on `PV.get_ctrlvars()` eventually having a valid value for precision and where `None` does signal "not known".    Should we merge this?

FWIW, I keep running into situations where `None` is used to signal either or both of "unknown" and "known to not exist".  It's almost as if there aren't enough singletons.  No, I think I didn't really mean that.  Maybe.

danielballan:
Haha, I hear you. Sometimes I think more sentinel values would make like easier and sometimes I think it would make it harder. Fine distinctions could go on without end: NaN, undefined, missing, empty, unknown....

@dchabot @klauer @tacaswell Is there any reason we couldn't start using `get_precision` in ophyd? If we can, then it sounds like this PR would solve our problem and be minimally disruptive.

newville:
OK,  I'll merge this.

PR #60, opened 2016-04-06
Add travis configuration for CI testing

klauer:
This adds preliminary continuous integration testing via a docker image that runs an EPICS IOC with a simulated area detector, simulated motors, and the pyepics testioc records.

That docker image is currently external, but it should probably be brought into the pyepics organization: 
* [github repo](https://github.com/klauer/simioc-docker/tree/pyepics-docker) (where the Dockerfile is located that generates the docker image)
* [docker hub](https://hub.docker.com/r/klauer/simioc-docker/) (where the docker image is stored)

Here's what a passing build looks like: https://travis-ci.org/klauer/pyepics/builds/121256923

~~TODO: change the currently-skipped tests which reference real motors to use simulated motors~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,90 @@
+language: python
+
+sudo: required
+dist: trusty
+
+services:
+  - docker
+
+cache:
+  directories:
+    - $HOME/.cache/pip
+
+matrix:
+  include:
+  # - python: 2.7
+    - python: 3.4
+  # - python: 3.5
+
+before_install:
+  - export DOCKER0_IP=$(/sbin/ifconfig docker0 |grep 'inet addr' | sed -e 's/.*addr:\([^ ]*\).*/\1/')
+  - export EPICS_CA_ADDR_LIST=$( echo $DOCKER0_IP | sed -e 's/^\([0-9]\+\)\.\([0-9]\+\)\..*$/\1.\2.255.255/' )
+  - export EPICS_CA_AUTO_ADDR_LIST="no"
+  - export EPICS_CA_MAX_ARRAY_BYTES=10000000
+  - export DOCKERIMAGE="klauer/simioc-docker"
+  - export DOCKERTAG="pyepics-docker"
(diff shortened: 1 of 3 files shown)


danielballan:
Excellent! If it passes, let's get this going. Some tests are always better than no tests.

tacaswell:
power cycled to kick travis

newville:
@klauer Thanks -- this looks good, but I know very little about how to make docker work, especially with travis.    

I don't really understand the problem... it's Python 3.5 only (but 3.4 is OK?) that fails.   But, it's associated with the darn 1 element char arrays!  Ack!

Any ideas what might be causing that?

klauer:
Hmm, this passed with my 3.5 tests [here](https://travis-ci.org/klauer/pyepics/builds/121262132). It may be a race condition with the monitor callback happening after these `get()` calls (which use the cached value).  

I'll tweak the test a bit with a delay and see if that helps. I'd normally fix such a sequential test by using `get(auto_monitor=False)`, but in this case we're actually testing the auto monitor unpacking.

klauer:
Fixed those tests, but there still is a not-insignificant chance that py.test itself will segfault after successfully running all of the tests. I had to re-run the 3.4 test after 5afc6bc to get it to pass.

klauer:
Continuing my comment spam...

Here's a gdb backtrace when travis fails: https://travis-ci.org/klauer/pyepics/jobs/121449424
I think we've seen this elsewhere, and it's a matter of pyepics not cleaning up nicely before exit. Should look into fixing that in the near future. I think this is ready otherwise.

Also @newville, docker/travis is difficult to get going from scratch, but with this PR as a template, tweaking it to do what you want shouldn't be so bad. If you have any questions as to what's going on in those files (as it's definitely not obvious), feel free to shoot me an e-mail.

newville:
@klauer looks great to me.  Any objections to merging?

newville:
@klauer @tacaswell Thanks!

PR #61, opened 2016-04-14
Fix bug in PV._set_charval when numpy is not installed

RobbieClarken:
When the PV class goes to generate the `char_value` for a char waveform it checks whether the value is a numpy `ndarray` and if so converts it to a list. This check was producing an `AttributeError` if numpy was not installed and thus failing to set `char_value`. This change makes `_set_charval` first check if numpy is available before doing the`isinstance` check.

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -360,7 +360,7 @@ class PV(object):
             return val
         # char waveform as string
         if ntype == dbr.CHAR and self.count < ca.AUTOMONITOR_MAXLENGTH:
-            if isinstance(val, ca.numpy.ndarray):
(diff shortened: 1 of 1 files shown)


newville:
@RobbieClarken Thanks!  I don't know why this is failing tests, but the failure doesn't look related to this PR.   @klauer has made great improvements to the testing code, and it looks like this failure is from a missing .travis.yml file, which must be a separate problem. 

So, I think it's safe to merge this change and figure out the testing stuff separately.

klauer:
I think it's because @tacaswell enabled travis on the repository but didn't disable it from running when `.travis.yml` is missing. Then because the travis PR hasn't been merged yet it failed.

PR #62, opened 2016-04-20
FIX: python 3 atexit cleanup

klauer:
I ~~think~~ hope this will finally resolve the crashes we've seen both locally and on travis.

This PR adds:
* atexit cleanup: `clear_channel` called on all PVs in the cache
* checks to make sure `PV.disconnect` doesn't call `clear_subscription` after the atexit code is called
* automatic gdb backtrace shown if py.test still segfaults (see 3.4 or 3.5 [here](https://travis-ci.org/klauer/pyepics/jobs/124517278) for an example)

Side-note: we need some multi-context tests to consider them supported


newville:
Great! Thanks!

PR #67, opened 2016-08-30
Timeout fixes

newville:
fixes for connection timeouts, especially so that unconnected PVs timeout in the specified time.


newville:
@klauer I don't understand this failure (Py3.4 only?) and don't see how the changes made could lead to this error.  Any ideas what is going on?

newville:
@klauer OK, and then it worked the second time.  Weird.  Is that a travis hiccup, or a subtle timing issue we should fix?

klauer:
@newville I think it's a timing issue that should be resolved at some point. I don't think your changes had anything to do with that, though.

Looks good to me.

newville:
OK, Thanks -  I think that connection timeouts cannot be related to that error about array values not matching.

Issue #72, opened 2016-11-10
restrict counts in callback of waveforms


aschaelicke:
Hallo,

we are using pyepics for monitoring large waveforms (25Msamples, 16bit, @ 1Hz or 20Msampels, 16bit @ 2Hz). In general we get good performance by activate auto monitoring and adding a callback to receive the data.
But in order to reduce the network load it was beneficial to tell the subscription how many samples are actually required.

In fact the *calib* already supports this via the *count* argument
```python
    ret = libca.ca_create_subscription(ftype, count, chid, mask,
                                       _CB_EVENT, uarg, ctypes.byref(evid))
```
So we modified  *ca.py* and *pv.py* to support this argument.

Would it be possible to add this to an official release?
[patch_pyepics3.2.6_count.txt](https://github.com/pyepics/pyepics/files/583273/patch_pyepics3.2.6_count.txt)

newville:
@aschaelicke this looks pretty straightforward.   I might nitpick that the default value for count better be "None" to signal "invalid" than -1.

Are you willing / able to make an actual Pull Request?   If not, I can apply these changes.

aschaelicke:
done in pull request #73

newville:
now fixed with #73

PR #73, opened 2016-11-15
allow restriction of counts in callback of waveforms

aschaelicke:
as discussed in issue #72

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -96,7 +96,7 @@ class PV(object):
                'upper_warning_limit', 'upper_ctrl_limit', 'lower_ctrl_limit')
 
     def __init__(self, pvname, callback=None, form='time',
-                 verbose=False, auto_monitor=None, count= -1,
+                 verbose=False, auto_monitor=None, count= None,
                  connection_callback=None,
                  connection_timeout=None):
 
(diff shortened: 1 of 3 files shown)


newville:
@aschaelicke This seems like a good idea to me.   I'd prefer the default value for `count` in `PV.__init__()` to be `None` instead of -1, and then use 
```python
     if self._args['count'] is None:
         self._args['count'] = count
     else:
         self._args['count'] = min(count, self._args['count']
```

It's not completely obvious to me what NELM should be in this case -- the "native length" on the server, or the maximum count for the subscription.  

It looks like several tests with waveforms fail with this change.  These might be cases where count=0 and/or 1, which can be tricky.  It might also be that there are assumptions about NORD and NELM that no longer hold when the subscription is restricted.  Anyway, these failures will have to be resolved, and ideally more tests added for this feature.

epics/pv.py
```diff
@@ -96,7 +96,7 @@ class PV(object):
                'upper_warning_limit', 'upper_ctrl_limit', 'lower_ctrl_limit')
 
     def __init__(self, pvname, callback=None, form='time',
-                 verbose=False, auto_monitor=None,
+                 verbose=False, auto_monitor=None, count= -1,
```
tacaswell: I would prefer `None` as a sentinal, not -1

TBirkeHZB:
We changed handling of optional count argument. 
Works properly now and passes all unittests - including the unittests we added for the new feature.

newville:
@TBirkeHZB  Thanks for the update and the test.  I like the separate "maxcount" too.

This looks good to me -- any objections to merging?

TBirkeHZB:
Not at all... And it would save Andreas (@aschaelicke) from applying his patch to every new version of pyepics in our local installations ;-) Thanks a lot!

newville:
OK, merged.  Thanks!

PR #74, opened 2017-01-17
update versioneer to 0.18

newville:
this should solve some known problems with older versions of versioneer and setuptools


tacaswell:
What sort of known problems?

newville:
@tacaswell  Just that older versions of versioneer made non-PEP440 names (https://github.com/warner/python-versioneer/blob/master/UPGRADING.md).   I was seeing hard-to-reproduce seg-faults from `python setup.py instlall` on some older linux systems, possibly with out-dated setuptools.   Upgrading versioneer seemed to help, and also avoids the non-PEP440 warnings.

PR #75, opened 2017-02-28
Fix puting single value to waveform of strings.

rokvintar:
Bug in ca.put() when working with waveforms of strings.

Old code is problematic since "cout" parameter is defined with "len(value)". In case of strings or bytes len() returns number of characters, but pypepics expects number of elements. When ftype == dbr.STRING, one string is one element, therefore count=1.

This results in following behaviour:
1) ca.put('abc')
- py2 .. instead of ['abc'], waveform will be ['a', 'b', 'c']
- py3 .. type error ... since value[i] is of type int if value is of type bytes

2) In second case when value is a list with a single element -->  ca.put(['abc'g), pyepics will try to handle it as a single value and will try to assign a list() to the object which expects strings(py2)/bytes(py3) ---> Type error.

This commit handles strings/bytes properly.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -1319,7 +1319,12 @@ def put(chid, value, wait=False, timeout=30, callback=None,
     if count > 1:
         # check that data for array PVS is a list, array, or string
         try:
-            count = min(len(value), count)
+            if ftype == dbr.STRING and is_string_or_bytes(value):
+                # len('abc') --> 3, however this is one element for dbr.STRING ftype
+                count = 1
+            else:
+                count = min(len(value), count)
+
             if count == 0:
                 count = nativecount
         except TypeError:
@@ -1333,9 +1338,9 @@ def put(chid, value, wait=False, timeout=30, callback=None,
     if is_string(value):
         value = ascii_string(value)
 
-    data  = (count*dbr.Map[ftype])()
+    data = (count*dbr.Map[ftype])()
     if ftype == dbr.STRING:
-        if count == 1:
+        if is_string_or_bytes(value):
             data[0].value = value
         else:
             for elem in range(min(count, len(value))):
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -292,10 +292,10 @@ class PV(object):
             # respect count argument on subscription also for calls to get 
             if count is None and self._args['count']!=self._args['nelm']:
                 count = self._args['count']
-                
             ca_get = ca.get
             if ca.get_cache(self.pvname)['value'] is not None:
                 ca_get = ca.get_complete
+
             self._args['value'] = ca_get(self.chid, ftype=self.ftype,
                                          count=count, timeout=timeout,
                                          as_numpy=as_numpy)
--- a/epics/utils3.py
+++ b/epics/utils3.py
@@ -50,4 +50,4 @@ def is_string_or_bytes(s):
     return isinstance(s, str) or isinstance(s, bytes) 
 
 def ascii_string(s):
-    return bytes(s, EPICS_STR_ENCODING)
+    return bytes(str(s), EPICS_STR_ENCODING)
--- a/tests/ca_type_conversion.py
+++ b/tests/ca_type_conversion.py
@@ -15,6 +15,7 @@ pvlist = (
     pvnames.long_arr_pv,       
     pvnames.double_pv,
     pvnames.double_arr_pv,
+    pvnames.string_arr_pv,
     )
 
 
--- a/tests/pv_type_conversion.py
+++ b/tests/pv_type_conversion.py
@@ -20,6 +20,7 @@ pvlist = (
     pvnames.long_arr_pv,       
     pvnames.double_pv,
     pvnames.double_arr_pv,
+    pvnames.string_arr_pv,
     )
 
 def onConnect(pvname=None,  **kw):
--- a/tests/pv_unittest.py
+++ b/tests/pv_unittest.py
@@ -204,8 +204,9 @@ class PV_Tests(unittest.TestCase):
         with no_simulator_updates():
             pv = PV(pvnames.string_arr_pv)
             put_value = ['a', 2, 'b']
-            with self.assertRaises(TypeError):
-                pv.put(put_value, wait=True)
+            pv.put(put_value, wait=True)
+            get_value = pv.get(use_monitor=False)
+            numpy.testing.assert_array_equal(get_value, ['a', '2', 'b'])
 
     def test_put_string_waveform_empty_list(self):
         write('String Array: put empty list\n')
--- a/tests/pv_unittest.py
+++ b/tests/pv_unittest.py
@@ -83,15 +83,6 @@ class PV_Tests(unittest.TestCase):
             self.assertIsInstance(val[1], str)
             self.failUnless(len(val[1]) > 1)
 
-    def test_put_string_waveform(self):
-        write('String Array: \n')
-        with no_simulator_updates():
-            pv = PV(pvnames.string_arr_pv)
-            put_value = ['a', 'b', 'c']
-            pv.put(put_value)
-            get_value = pv.get(use_monitor=False, count=len(put_value))
-            numpy.testing.assert_array_equal(get_value, put_value)
-
     def test_putcomplete(self):
         write('Put with wait and put_complete (using real motor!) \n')
         vals = (1.35, 1.50, 1.44, 1.445, 1.45, 1.453, 1.446, 1.447, 1.450, 1.450, 1.490, 1.5, 1.500)
@@ -190,6 +181,49 @@ class PV_Tests(unittest.TestCase):
         self.failUnless(len(NEWVALS) > 3)
         mypv.clear_callbacks()
 
+    def test_put_string_waveform(self):
+        write('String Array: put\n')
+        with no_simulator_updates():
+            pv = PV(pvnames.string_arr_pv)
+            put_value = ['a', 'b', 'c']
+            pv.put(put_value, wait=True)
+            get_value = pv.get(use_monitor=False)
+            numpy.testing.assert_array_equal(get_value, put_value)
+
+    def test_put_string_waveform_single_element(self):
+        write('String Array: put single element\n')
+        with no_simulator_updates():
+            pv = PV(pvnames.string_arr_pv)
+            put_value = ['a']
+            pv.put(put_value, wait=True)
+            get_value = pv.get(use_monitor=False)
+            self.failUnless(put_value[0] == get_value)
+
+    def test_put_string_waveform_mixed_types(self):
+        write('String Array: put mixed types\n')
+        with no_simulator_updates():
+            pv = PV(pvnames.string_arr_pv)
+            put_value = ['a', 2, 'b']
+            with self.assertRaises(TypeError):
+                pv.put(put_value, wait=True)
+
+    def test_put_string_waveform_empty_list(self):
+        write('String Array: put empty list\n')
+        with no_simulator_updates():
+            pv = PV(pvnames.string_arr_pv)
+            put_value = []
+            pv.put(put_value, wait=True)
+            get_value = pv.get(use_monitor=False, as_numpy=False)
+            self.failUnless('' == ''.join(get_value))
+
+    def test_put_string_waveform_zero_length_strings(self):
+        write('String Array: put zero length strings\n')
+        with no_simulator_updates():
+            pv = PV(pvnames.string_arr_pv)
+            put_value = ['', '', '']
+            pv.put(put_value, wait=True)
+            get_value = pv.get(use_monitor=False)
+            numpy.testing.assert_array_equal(get_value, put_value)
 
     def test_subarrays(self):
         write("Subarray test:  dynamic length arrays\n")


newville:
@rokvintar  Hm, sorry for the trouble, but I don't really follow your report.   
    Putting a Python string/bytes to a string PV works. This is reasonably well tested.
    Putting a Python string or a sequence of short integers to a character waveform PV also works (or at least, has a meaningful interpretation give the nature of strings and waveform PVs).
    Putting a list with one element to a string PV gives a TypeError.
    Putting a list of one-element strings to a character waveform does not work.  

I think you might be suggesting that one of the latter two be changed.  

A PR should solve a demonstrated, documented problem. It should provide actual code and a full description of the problem it produces, including the full traceback if one is produced.   A PR should also add a test for the problem or explain why it does not.

I find that often times people raise issues or submit PRs when they really want to ask a question about usage.  So: what is it you are trying to do?

rokvintar:
Hi sorry for bad explanation.
The problem is not with waveform of chars, but with waveform where FTVL field is "STRING".

Here is an example for python 3
```
import epics
p = epics.PV('TST:wfStr')
# Following works
p.put(['abc', 'bcd'])

# Following does not
p.put(['abc'])       
```
It fails with:
```
  File "<stdin>", line 1, in <module>
  File "/home/rvintar/workspace/CSL/SaveRestore/pyepics/pyepics/epics/pv.py", line 356, in put
    callback_data=callback_data)
  File "/home/rvintar/workspace/CSL/SaveRestore/pyepics/pyepics/epics/ca.py", line 419, in wrapper
    return fcn(*args, **kwds)
  File "/home/rvintar/workspace/CSL/SaveRestore/pyepics/pyepics/epics/ca.py", line 1339, in put
    data[0].value = value
TypeError: bytes expected instead of list instance
```
------
Why this happens:
In ca.py line 1322:
```
count = min(len(value), count)
```
In case of ['abc']  len(value) returns 1.

And then in lines 1338  and 1339... if count ==1 (which it is) pyepics wants to write a list to the first element of waveform of strings (bytes expected).

I hope this helps.

Kind regards,
Rok

newville:
@rokvintar  Hm, that is not the behavior I see.   I get the same failure of "TypeError: bytes expected instead of a list instance" (or 'string' instead of 'bytes' for Python 2) when putting a list containing one or two strings to a string PV:

```Python
>>> import epics
>>> p = epics.PV('Py:str1')
>>> p
<PV 'Py:str1', count=1, type=time_string, access=read/write>
>>> p.put(['abc', 'def'])
  File "<stdin>", line 1, in <module>
  File "/home/newville/Codes/pyepics/epics/pv.py", line 356, in put
    callback_data=callback_data)
  File "/home/newville/Codes/pyepics/epics/ca.py", line 419, in wrapper
    return fcn(*args, **kwds)
  File "/home/newville/Codes/pyepics/epics/ca.py", line 1339, in put
    data[0].value = value
TypeError: bytes expected instead of list instance
>>> p.put(['abc'])
  File "<stdin>", line 1, in <module>
  File "/home/newville/Codes/pyepics/epics/pv.py", line 356, in put
    callback_data=callback_data)
  File "/home/newville/Codes/pyepics/epics/ca.py", line 419, in wrapper
    return fcn(*args, **kwds)
  File "/home/newville/Codes/pyepics/epics/ca.py", line 1339, in put
    data[0].value = value
TypeError: bytes expected instead of list instance
```

That doesn't seem like a problem to me.  I would expect to put a Python bytes or string object to a string PV.   My inclination would be to have a TypeError when putting any list to a string PV.  Why should that ever "work" (and what does "work" mean)?

Again, it would be helpful to enter the conversation as "this is what I want to do" rather than as "this is the solution to the problem".  What are you trying to do?  If it is "put any data structure to a string PV", you will be able to find a lot of ways for that to not work. 

Do you actually want to put a list of strings to a string PV?  What are you expecting the result of that to be?

rokvintar:
@newville As I have explained, I am need to put to a **waveform of strings** (not waveform of chars, nor a string record).

```
p = epics.PV('TST:wfStr')
p
<PV 'TST:wfStr', count=0/200, type=time_string, access=read/write>
```

newville:
@rokvintar Ah, sorry for not understanding your data type.  I don't think I've ever encountered a waveform of strings.  Apparently they aren't used much, though I don't know why that is....  Anyway, as you no doubt understand, we don't currently have any tests with waveforms of strings. 
 
While I can certainly believe your fix is perfect, we have made made any decisions about how to deal with these. Most waveforms are mapped to numpy ndarrays, but I doubt you would really want an object ndarray -- you would probably want a list of strings.  That might need special handling. 

We also are not testing waveforms of strings in any of the test suite.  If we're going to claim that pyepics works for these, we should probably decide how they should be handled, and they should be added to the unit tests, including the type conversion routines.

rokvintar:
@newville I see. I know this is not a type that is commonly used, but in cases of applications like save and restore this would mean that users are somehow prohibited to use waveforms of strings.

Regarding the expected type ... I don't really care if it is a list or ndarray of strings. Currently pypeics seems to return a list of strings ... and this is fine with me.

-------
I have one more question regarding general work with waveforms.
For example: I have a waveform of longs and it's current value is array with only one element. For example: [2]. What should I expect when doing pv.get():
a) a single value 2
b) or numpy.array([2])

```
import epics
p = epics.PV('TST:wf')
p
<PV 'TST:wf', count=0/200, type=time_long, access=read/write>

# Single value in array
p.put([2])
p.get() # Here I would expect: array([2], dtype=int32) like in cases with multiple vales
>>> 2

# Multiple values in array
p.put([2,3,4])
p.get()
>>> array([2, 3, 4], dtype=int32)
```
This is again case which is not really common in practice, but I would expect that if I have a waveform record, pv.get() will always return an array, regardless of number of current elements.

What do others think?

newville:
@rokvintar  Does autosave use waveforms of strings?  I wasn't aware of that.  

Anyway, I'm OK with supporting string waveforms, but in order to do so we need actual tests.   As this is a PR,  it should include tests for working with string waveforms, including normal type conversions and also handling the corner cases of one-element arrays, zero-length strings, what to with a list of mixed types, and so on. 

For your other question: 1-element arrays cause many troubles in lots of situations and codes.  I'd be reluctant to change the current behavior without ample discussion and testing of both pyepics and downstream apps.

rokvintar:
@newville
1. I will provide unit test and normal type conversions but it might take some time to do it.
2. Regarding 1 element array. I just wondered if anyone else is aware of this possible bug since it means that all applications that expect waveforms with 1 element must handle this case separately.
3. I don't know if autosave part of pyepics handles waveforms of strings, but save and restore application ([snapshot](https://github.com/paulscherrerinstitute/snapshot)) developed at Paul Scherrer Institute does.

rokvintar:
Dear @newville

I have added the tests for waveforms of strings. Please check, and if necessary I can add/change something.

Kind regards,
Rok

newville:
@rokvintar Thanks!   It looks like the test with mixed types (strings and ints) is giving differing results.   I think that for Python2.7 the put() will succeed, as the put essentially puts a str() around each value.  Not sure why Python3.4 and Python3.5 are different from each other, though -- I think they should both give TypeErrors, but only because `bytes(integer_value)` fails.

I think my suggestion would be to not be so strict on what, if any, exceptions might occur in this situation:  That is, should putting "['a', 'b', 3]" to a string waveform be an error at all?  If so, should that error by a TypeError?  

I'm not sure about replacing the test of "count==1" with "is_string_or_bytes"  in ca.put() for ftype=dbr.STRING is really the right thing to do.  That is, I'm not sure I really have an opinion on what 
```
    p.put('a single string')
```
should do for a string waveform.   The meaning is ambiguous.  Making a defensible guess is fine, as is raising an error. 

That said, I'm also not sure that replacing "count==1" with "is_string_or_bytes" is wrong, or worse.  String coercion is always sort of a mess, as is handling 1-element arrays.   Here we have both.    

Does anyone have an opinion on what the behavior for these cases should be?

newville:
@rokvintar Hi, how is this PR coming along?  It seems like (after overcoming my own density) this was pretty complete, maybe with a few not-completely-clear corner cases (putting a single string to a string waveform: that should that go to the first element, but it's kind of ambiguous) and some questions about what had been `count==1` cases.

rokvintar:
@newville I have been on vacation and then forgot a bit about this issue. In the following days I'll take a look of it again to see what can be done to cover the edge cases, to make this fix right.

gabrielfedel:
Hi, any news about this PR?
I'm having problems with pyepics 3.2.5 and 3.2.6 related to this. Thank you!

newville:
@gabrielfedel  @rokvintar  I agree that this would be good to merge, but we'd vastly prefer to merge changes that pass all tests.

I think that the failure here is really with a test that is expected to fail, but that raises a ValueError with Python2 and a TypeError with Python3 (I could be wrong about that).   If that is right, fixing the failing test should be easy to do.   Are either of you up for doing that?

rokvintar:
Hi,
 
I took a look at it and following is what I have figured it out:
1) Mixed types:
put(['a', 2, 'b'])

Python 2: This put was successful since utils2.py defines:
`ascii_string = str`

Python 2: It fails when it tries to map int to bytes since utils3.py defines
```
 def ascii_string(s):
         return bytes(s, EPICS_STR_ENCODING)
```

What I see as an option:
- allowing mixed types would be pythonic, therefore ['a', 2, 'b'] is treated as a valid case
- to allow it implementation of `ascii_string` in utils3.py should be changed to:
```
 def ascii_string(s):
         return bytes(str(s), EPICS_STR_ENCODING)
```
- Mixed types testcase can then  be changed to expect that put() succeeds. Does not wait for "Type Error" any more. 

This changes will force pyepics to behave same way for py2 and py3 whenever ascii_string() is used which by my opinion is a good and desired thing.

2) Then questions regarding the `count==1` cases:
ca.put() must be able to recognize cases where a count>1 does not mean there is an array or sequence of values, but a single string which is one element for the waveform of strings. This is why the implementation is made this way.

newville:
@rokvintar ah, yeah.  Now I remember why this is so ugly ;).  If you have a waveform of strings and put a list or random types, what should happen?  Maybe
```
   put([ascii_string(s) for s in input_list])
```
is the most sensible thing to do?  But `input_list` should really be any iterable object.  Oh, except strings or bytes. 

Hm, maybe it's just me, but I keep thinking that the main problem is that we don't have a solid concept of what /should/ happen in all these cases.   Do you have an opinion?

rokvintar:
@newville What you wrote is exactly the behavior that is currently implemented for py2, and would be also for py3 if  `ascii_string(s)` is changed according with the above suggestion (at least I cannot find a difference). I think, generally this is a good approach for mixed types arrays for waveform of strings.

newville:
@rokvintar  I sort of think that the current PR  is more correct in Py3 where it raises an error and refusing to guess how to coerce the integer 2 into a string.  

But, as you say, that coercion is normal for Py2.   So, perhaps the best approach is to change the test to expect `put(['a', 2, 'c'])`  to work in Py2 and fail in Py3 with a TypeError.  That is, not changing the behavior but the expected-ness of the current result.

And, just to be clear and because I'm slightly flaky about such things (and to your point 2 above): I think this PR is doing the right thing, even in the tricky intersection of "Arrays with 1 element" and "Strings are sort of like sequences of bytes/characters and sort of like single objects".

rokvintar:
@newville By my opinion we must correct current PR with some additional changes. I would says that first and golden rule for the library such as pyepics should be that it behaves the same way in py2 and py3. If you agree with it then this lives us 2 options:
1. Do it like current PR does it for py3
2. Do it like current PR does it for py2

Independent from the above decision you must still consider that `ascii_string(s)` should also behave same ways in py2 and py3 ... which is currently not the case.

Maybe @gabrielfedel  has an opinion?

gabrielfedel:
I think I agree with @rokvintar . It looks like the best thing is the behaviour be the same for py2 and py3. 

Maybe the option 1 ( Do it like current PR does it for py3 ) looks better (so, probably will need to check the type on py2, right ?)

If you agree with this solution, I can propose some code for it.

rokvintar:
For me either option 1 or 2 is acceptable.
For option 2 I already have an implementation.
For option 1 I would be glad if @gabrielfedel can take over.

newville:
@rokvintar  That's a good point:  we are already converting the Python types to Epics types and do plenty of type coercion in those steps (both ways).  So I agree that we should make the Py3 behavior match the Py2 behavior.  so `bytes(str(s), EPICS_STRING_ENCODING)` seems like the right thing to do. 

Changing something as low-level as `utils3.ascii_string` is a little scary, but that's what we have tests for, right? ;)

rokvintar:
I have pushed changes of the `ascii_string(s)`. Also the unittest was adopted to this behavior. @newville  can you please check and merge if it is ok.

This change should not break any other functionality. As far as I know it is currently used only by ca.put

newville:
@rokvintar hm, it seems weird that this is still failing tests.  The code looks OK to me. Do you know what is happening? I'll try to track this down, but it might take me some time....

rokvintar:
@newville It's true, it seems that some tests are failing. I don't see which code changes would cause the fails.

newville:
@rokvintar I'm merging this to master now, and then will try to better understand and fix the errors.  

Master may be broken for a short time, but once fixed, I think it will be ready to be the next release.

Issue #76, opened 2017-04-03
PV Access Rights are not updated dynamically, out of sync with server's state 


dchabot:
In an environment where channel access rights (read/write) can change dynamically, _pyepics_ does not allow for this, and `PV` objects will always convey whatever their read/write access rights were [on connection](https://github.com/pyepics/pyepics/blob/master/epics/pv.py#L186). Note, a `PV` that has had its read/write access revoked will still `raise` if a read/write is attempted.

Access Rights can change dynamically due to CA Security rules going in and out of effect, for example (guess how I found this problem??).

See [test case](https://gist.github.com/dchabot/3f63bd0f6c7b7e1644aeabaa82b03f5b)

I would suggest that the way to deal with this is to have all `PV` objects register an [_access_rights_handler_](http://www.aps.anl.gov/epics/base/R3-14/12-docs/CAref.html#ca_replace).

newville:
@dchabot  Yep, this is a problem.    We see this too, and not just with Python clients ;).  

I agree we should expose the access_rights handler... I think I didn't know this existed!  It's not even mentioned in ca.py, though other crazy things like the printf handler and all the `sg_*` functions are there...  

Anyway, yes, we should add this and use it.

dchabot:
I'll start a PR for this.

dchabot:
Closed by #77

PR #77, opened 2017-04-04
ENH: access rights event handling

dchabot:
Addresses #76.

Todo:
- [x] basic access rights event handling
- [x] docs
- [x] tests
- [x] API changes for `PV` class

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ script:
   - caget Py:long1
 
   # running tests
-  - py.test -vv --cov=epics --cov-report term-missing tests/ca_unittest.py tests/pv_unittest.py tests/test_pool.py tests/test_threading.py || RESULT=$?
+  - py.test -vv --cov=epics --cov-report term-missing tests/ca_unittest.py tests/pv_unittest.py tests/test_pool.py tests/test_threading.py tests/test_cas.py || RESULT=$?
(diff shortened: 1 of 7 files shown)


klauer:
In my opinion, it would be sufficient to run callbacks such that any that care about it could check the access rights. It seems closer to the metadata that comes along with `DBR_CTRL`/`DBR_STS`, even though in the CA protocol it comes from a completely different command header.

(Only my 2c as I no longer really have any application for any of this...)

newville:
@dchabot  Thanks very much! That looks very good on first read and without trying it.  I think I agree that this should be part of every PV, or at least on by default. 

Out of curiosity, do you know when the callback first gets called in the connection process?  Like, for how long will the access fields (`self._args['read_access']`, etc)  be `None`?  It's probably not a big deal.

Perhaps the next question is whether we should use this information to raise an error or warning if a put() or get() would violate the current access rights. 

I think  a user-supplied callback would be useful. For example, you want a GUI to be able to change color or some other property to indicate that a PV not writable.  Most of the display managers do this. I use such callbacks to indicate connection state and I think it would be useful to show when a PV has gone into "read-only" mode.

newville:
@dchabot Thanks very much!   I agree that it looks good to go, and  I'm eager to try it out.    Since this passes all tests and is mostly (solely?) providing new features, my inclination would be to merge this.  

Anyone have any objections, concerns, or comments?

newville:
OK, merging this,.  Thanks @dchabot !

PR #78, opened 2017-04-10
CI: tweak how docker is started

tacaswell:
This will make it possible to start more than one docker instance
and have them live side-by-side.

This binds each servers 5064 tcp port to a random port in the range
7000-8000 which is sorted out by libca as part of the search process.

The udp multi-cast passes through on 5064 (as it did before).

We do not need to expose the heart-beat outside of the docker manager.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
   - docker pull ${DOCKERIMAGE}:${DOCKERTAG}
   - docker images
   - docker ps -a
-  - docker run -d -p $DOCKER0_IP:5064:5064/tcp -p $DOCKER0_IP:5065:5065/udp --name epics_iocs ${DOCKERIMAGE}:${DOCKERTAG}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ before_install:
   - docker pull ${DOCKERIMAGE}:${DOCKERTAG}
   - docker images
   - docker ps -a
-  - docker run -d -p $DOCKER0_IP:5064:5064/tcp -p $DOCKER0_IP:5065:5065/udp --name epics_iocs ${DOCKERIMAGE}:${DOCKERTAG}
+  - docker run -d -p $DOCKER0_IP:7000-8000:5064/tcp -p --name epics_iocs ${DOCKERIMAGE}:${DOCKERTAG}
```
dchabot: @tacaswell - copy/paste error(s) here?
This would never work as is...

tacaswell: deletion failure :(

newville:
@tacaswell Thanks -- this looks good to me.   Anyone have any comments/concerns?  Otherwise, I'll on merging in a day or so.

tacaswell:
Running the docker image we use for ophyd testing along side this docker image is how I started down this path.

PR #80, opened 2017-04-11
Include libca dlls

newville:
This adds pre-built dynamic libraries for libca and libCom for darwin (64 bit only) and linux (32 and 64 bit) as well as windows 32 and 64 bit.  

These will be installed as data with the library, according to platform details.  

The linux64 libraries are built without readline support, so as to avoid conflicts with Anaconda Python.  The linux32 libraries are fairly old, and seem to have been built against readline 5.... should probably investigate whether these should be updated.  Ideally, all these libs should be freshly built with 3.15 -- I have not done that yet on all platforms.

With this change, the "find_libca()" process is:
   1.  Look for PYEPICS_LIBCA, use if available.
   2. Look for these "installed-with-pyepics" versions.
   3. Search PATH, then "likely install locations".   

I think that is unlikely to break any existing installations, and should still work on non-linux Unixes, though I don't know where those are actually used.

I'm in no rush to merge this, and want to test more thoroughly with Windows, but it seems like something we ought to do (if for no other reason than to avoid the dreaded readline conflicts with Anaconda Python).   Suggestions, comments, better ideas would be much appreciated!


newville:
had another question about pyepics on linux+Anaconda, so I'm going to merge this and we can fix any problematic or out-dated dlls as needed.

Issue #81, opened 2017-04-20
libca/libCom readline conflicts on Linux, especially Anaconda


newville:
I recently pulled into master #80 -- an attempt to distribute libCom and libca for more platforms (linux, darwin) that are *not* built with readline support.  This is especially to avoid conflicts between system readline and Anaconda readline.

The PR has a problem for linux64 that libca.so looks for "libCom.so.3.15.5" -- this can be fixed with a symlink.

**But** it also has a larger issue in that the install with setuptools installs these libs inside the "pyepics....egg" and the libca library still cannot find libCom.so.*.  

I believe the solution is to revert from setuptools to distutils.core so that the installation from the source kit goes to {ANACONDA}/lib/pythonX.Y/epics, and the "data" dlls go to {ANACONDA}/lib.   I would be very happy to give up the eggs thing.

Does this seem reasonable to others?   If not, how do others handle running on linux with anaconda?

tacaswell:
You should have a look at how the manylinux wheels are built bundle the libraries that way.  They have done the hard work of resolving all of the library dependencies and bundling them up.

https://github.com/pypa/manylinux

https://github.com/MacPython/matplotlib-wheels

https://github.com/MacPython (anything that ends in -wheels)

dchabot:
Is this the conda readline issue complaining about the "unknown PC symbol"?

I think reverting to `distutils.core` is a step backwards.
Losing developer-installs, `python setup.py develop`, would be a pain.

newville:
@tacaswell  Isn't that intended to solve the problem of variations in linux library versions? I could be wrong, but I think that's a separate issue.  There may be linux-to-linux conflicts, though I don't know that there are.  

The problem here is not that the dynamic libraries don't load, but that they aren't *found* because they are help inside the pyepics...egg file.

That said, I'd be OK with building the epics base on an older Linux platform, but haven't looked into manylinux in detail.

@dchabot  losing `python setup.py develop` would be a downside -- I tend to use separate environments so don't use that feature much myself any more, but it's a nice option.  
OTOH, not finding library files that are included in the package is a downside too.  To be clear, the shared object files are in pyepics....egg/lib/ and `epics.ca.find_libca()` finds the `libca.so` there, but loading that fails to load the `libCom.so.3.15.5` in the same folder.   Even if  I set LD_LIBRARY_PATH to that directory.  That's sort of a problem ;).  Setuptools is not without issues. 

[...]

PR #82, opened 2017-04-24
set RPATH and SONAME for linux64 shared libraries

newville:
Following suggestions from @dchabot, this PR addresses #81 by altering the linux64 shared libraries to work when embedded in the pyepics "egg" from setuptools.

The patch was made with (in `dlls\linux64`):
```
patchelf --set-soname libca.so  libca.so
patchelf --set-soname libCom.so  libCom.so
patchelf --set-rpath '$ORIGIN' libca.so
patchelf --set-rpath '$ORIGIN' libCom.so
patchelf --replace-needed libComs.so.3.15.5 libCom.so libca.so
```


ralphlange:
I think that with a properly configured EPICS Base build, none of the above post-build commands are necessary. Both RPATH and SONAME can be set in the EPICS build, and the resulting libraries can be used as they are.
Of course, technically this is equivalent to changing these settings afterwards.

newville:
@ralphlange  Yes, I agree that would be preferable.  We have to alter the configuration to turn off readline, so we should probably set this up too.  

Maybe even record the detailed changes to make it easier to build working dlls next time!

newville:
@dchabot  No problem -- my view is that this is definitely a conversation, and I am trying to not have any strong opinions on any of this topic.  I *did* merge to master a preliminary idea to make it easier for someone else to test... the idea is incomplete, and the result didn't work due to the names and paths of the shared objects. 

You're right that the search order for libca changed.   My thought was to move toward including shared libraries that we "know work" (or at least, that work for a great many machines) and would not require otherwise installing epics base.   This approach is definitely needed for Windows, is at least arguably "nice" for OSX, and could easily be viewed as "useful" on Linux.  With the whole Anaconda readline issue, this seems more than just "useful" on Linux -- either we provide a shared library or have some other way of distributing non-readline versions of libCom.

If we do this (and if it works!), then we *could* get rid of the search options 3 and 4, so that the search becomes "use locally provided libca unless PYEPICS_LIBCA" is set.   That could be viewed as more reliable (wrt to Anaconda+readline), and would mean that a pyepics package would not need epics base installed, which might be useful for deployment to many client machines, especially in considering that they might not otherwise need a compiler chain installed (OSX, Windows, raspberryPI).

ralphlange:
One thing to keep in mind:
When using properly packaged EPICS, "installing EPICS Base" is a minimal, streamlined system operation (that +- installs a package with a few libraries).
As an administrator of an infrastructure that uses packaged EPICS, I will have a strong interest that all applications on the system use my package-installed EPICS libraries, and applications that bring their own embedded copies of my packaged libraries are a big pain and continuing source of incompatibility troubles and maintenance issues.
On the other hand, packaged EPICS libraries will (hopefully!) always have proper SONAME settings and no RPATH, which doesn't work well with the portable concept of pyepics binding to the EPICS libraries. So, package using environments will probably have to create pyepics packages that depend on, work together with and use the EPICS packages.

newville:
@ralphlange Thanks for the comments.

On Tue, Apr 25, 2017 at 12:20 PM, Ralph Lange <notifications@github.com>
wrote:

> One thing to keep in mind:
> When using properly packaged EPICS, "installing EPICS Base" is a minimal,
> streamlined system operation (that +- installs a package with a few
> libraries).
>

That's fine. Out of curiosity, what packages / package systems?  And should
pyepics *rely* on these?  I'd rather not have to restrict supported use of
pyepics too much, say by requiring Anaconda or a certain flavor of Linux,
but having coordinated packages for those sort of systems would definitely
be useful.

> As an administrator of an infrastructure that uses packaged EPICS, I will
> have a strong interest that all applications on the system use my
> package-installed EPICS libraries, and applications that bring their own
> embedded copies of my packaged libraries are a big pain and continuing
> source of incompatibility troubles and maintenance issues.
>

Again out of curiosity (really), why would it be a problem if a CA client
library for language X installed its own version of libca that was hidden
away in some package-specific directory?  The client side of CA seems to
move pretty slowly to me.  We often have a range of libca clients at our
beamlines.

But, if it is a problem, would setting PYEPICS_LIBCA to force the libca
used a reasonable solution?

On the other hand, packaged EPICS libraries will (hopefully!) always have
> proper SONAME settings and no RPATH, which doesn't work well with the
> portable concept of pyepics binding to the EPICS libraries. So, package
> using environments will probably have to create pyepics packages that
> depend on, work together with and use the EPICS packages.
>

My concern for using such packaged libraries is that the packaged libCom.so
would almost certainly be linked against readline, which means it may not
work with other applications.   To be clear, I think having libCom.so link
to readline is the real problem here, but I'm not expecting to win that
battle, and trying to come up with a viable alternative.  Including binary
libca/libCom is one option for that.

I'm not sure why packages with environments would necessarily have to do
anything unusual for a pyepics package.  Under the current proposal,
installing pyepics would put libca/libCom in a place that really only
pyepics would be expected to find them, under
pythonX.Y/site-packages/pyepics.....egg/lib.   And you could set
PYEPICS_LIBCA to override that.

ralphlange:
>
> > One thing to keep in mind:
> > When using properly packaged EPICS, "installing EPICS Base" is a minimal,
> > streamlined system operation (that +- installs a package with a few
> > libraries).
>
> That's fine. Out of curiosity, what packages / package systems?


All of them, of course!
At least the mainstream Debian (.deb) and RHEL/CentOS (.rpm) worlds, maybe
Chocolatey on Windows?

And should pyepics *rely* on these? I'd rather not have to restrict
> supported use of
> pyepics too much, say by requiring Anaconda or a certain flavor of Linux,
> but having coordinated packages for those sort of systems would definitely
> be useful.
>

Yes, exactly.
pyepics (upstream) should *not* rely on these, but packaged pyepics (.deb
.rpm) should.
https://github.com/epicsdeb is a start in that direction, and once we start
adding RPM packages, it should be done in the same packaging repo, using
the same layout etc.

> As an administrator of an infrastructure that uses packaged EPICS, I will
> > have a strong interest that all applications on the system use my
> > package-installed EPICS libraries, and applications that bring their own
> > embedded copies of my packaged libraries are a big pain and continuing
> > source of incompatibility troubles and maintenance issues.
>
> Again out of curiosity (really), why would it be a problem if a CA client
> library for language X installed its own version of libca that was hidden
> away in some package-specific directory?


Duplication of code, inconsistent behaviour between different clients that
is hard if not impossible to debug.

The client side of CA seems to move pretty slowly to me. We often have a
> range of libca clients at our
> beamlines.
>

True, but libCom changes pretty drastically, and has seen e.g. inside the
3.15 series quite a lot of changes and bug fixes.

But, if it is a problem, would setting PYEPICS_LIBCA to force the libca
> used a reasonable solution?
>

Maybe. Environment variables are quite volatile and not really well
traceable, so personally I would try to avoid it.

@dchabot: I know you did that packaging... thank you for the effort.
Btw: I started - as a toy project with truly minimal progress so far - an
effort to set up an "official" EPICS packaging infrastructure, to create
packages that would be hosted publicly in repositories on the future EPICS
web site. The system I am looking at would be able to build for multiple
Linux distros / packaging systems.

newville:
On Tue, Apr 25, 2017 at 2:38 PM, dchabot <notifications@github.com> wrote:

> I don't want to speak for @ralphlange <https://github.com/ralphlange>,
> but I've packaged pyepics for debian systems
> <https://github.com/epicsdeb/pyepics>, for example. Admittedly, this was
> abandoned in favor of conda for that facility (nsls2), but I am pursuing
> this again for use at FRIB, where one of the few vectors available for code
> deployment is the debian packaging system.
>
> One possible runtime issue with systems that use multiple versions of the
> same library (say, libca.so.x.y and libca.so.x.z) occurs when trying to
> track down problems ("is it server-side libca.so.x.y or client-side
> libca.so.x.z?").
>

But isn't that true in all cases?  There is a server that uses a version of
libca.  There is a client that uses a version of libca.  They aren't the
same.

> Another issue your proposal raises is coupling libca version to pyepics.
> This is something to be avoided, as it breaks ease of upgrading (or
> downgrading!) library versions.
>

Huh?  How does it break ease of upgrading library versions?    If
libca/libCom gets installed with pyepics in the python
site-package/pyepics.../ directory, you can easily alter other versions of
libca/libCom without affecting the pyepics version.   This is more or less
the classic solution to "dll hell" -- distribute per-app versions of the
library, squirreled away where they are unlikely to be found accidentally
by another app.

If a user wants to use a different version of libca, they can.

@newville <https://github.com/newville>
>
> But, if it is a problem, would setting PYEPICS_LIBCA to force the libca
> used a reasonable solution?
>
> Please don't do that. I think my suggestion above (re-ordering the libca
> searching) gets us both what we're looking for and is backwards compatible.
>

What's the objection to an environmental variable?  My experience is that
you basically have to have several epics-specific environmental variables
set for epics to work correctly.

Your proposed ordering does not solve the Anaconda+readline problem.  That
is, if a system libCom.so is linked to libreadline, there is a very good
chance that it will not work with Anaconda Python.  This is the problem
that started this topic, and it has to be solved.  If we do not distribute
and install libraries pre-compiled without readline, there needs to be
another way to solve this problem.   Do you have a different suggestion?

newville:
On Tue, Apr 25, 2017 at 2:49 PM, Ralph Lange <notifications@github.com>
wrote:

> >
> > > One thing to keep in mind:
> > > When using properly packaged EPICS, "installing EPICS Base" is a
> minimal,
> > > streamlined system operation (that +- installs a package with a few
> > > libraries).
> >
> > That's fine. Out of curiosity, what packages / package systems?
>
>
> All of them, of course!
> At least the mainstream Debian (.deb) and RHEL/CentOS (.rpm) worlds, maybe
> Chocolatey on Windows?
>
> And should pyepics *rely* on these? I'd rather not have to restrict
> > supported use of
> > pyepics too much, say by requiring Anaconda or a certain flavor of Linux,
> > but having coordinated packages for those sort of systems would
> definitely
> > be useful.
> >
>
> Yes, exactly.
> pyepics (upstream) should *not* rely on these, but packaged pyepics (.deb
> .rpm) should.
> https://github.com/epicsdeb is a start in that direction, and once we
> start
> adding RPM packages, it should be done in the same packaging repo, using
> the same layout etc.
>
>
OK, we agree that pyepics should not be limited to certain packages.  FWIW,
there is no way I am going to change pyepics to not install binaries of
ca.dll and Com.dll on Windows.  I'm actually kind of amazed to hear the
objections to distributing shared libaries, but it definitely seems you and
@dchabot are thinking mostly about Linux.



> > As an administrator of an infrastructure that uses packaged EPICS, I will
> > > have a strong interest that all applications on the system use my
> > > package-installed EPICS libraries, and applications that bring their
> own
> > > embedded copies of my packaged libraries are a big pain and continuing
> > > source of incompatibility troubles and maintenance issues.
> >
> > Again out of curiosity (really), why would it be a problem if a CA client
> > library for language X installed its own version of libca that was hidden
> > away in some package-specific directory?
>
>
> Duplication of code, inconsistent behaviour between different clients that
> is hard if not impossible to debug.
>


Do you really see this with CA clients and find that the inconsistencies
come from different libca libraries that cannot be easily identified by
version number?  I don't think I've ever seen that.



> The client side of CA seems to move pretty slowly to me. We often have a
> > range of libca clients at our
> > beamlines.
> >
>
> True, but libCom changes pretty drastically, and has seen e.g. inside the
> 3.15 series quite a lot of changes and bug fixes.
>

I'm not sure how much of libCom is actually used by libca, but this entire
problem is trying to address a problem with libCom (that it is usually
built with readline on linux) that would otherwise have no effect on libca.



>
> But, if it is a problem, would setting PYEPICS_LIBCA to force the libca
> > used a reasonable solution?
> >
>
> Maybe. Environment variables are quite volatile and not really well
> traceable, so personally I would try to avoid it.
>

How do you use EPICS clients that work with large data?  How do you deal
with display managers finding files?  Really, I'd like to hear answers to
those because for me, EPICS largely rests on setting environmental
variables.   I don't know how one can possibly build EPICS tools without
setting environmental variables, and I basically cannot use EPICS without
setting EPICS_CA_MAX_ARRAY_BYTES.  So, it's kind of astonishing to hear
that from both of you.

What would your suggestion be for distributing pyepics in a way that could
be used with Anaconda Python on Linux if we did not include libca/libCom
built without readline support and did not use any environmental
variables?  I know that does not exist right now, and I don't really have a
good idea what that would look like.  OTOH, a way to distribute pyepics so
that it can work on Linux with Anaconda or (probably) any other version of
Python is ready to be merged.

ralphlange:
On Wed, Apr 26, 2017 at 12:25 AM, Matt Newville <notifications@github.com>
wrote:

>
> OK, we agree that pyepics should not be limited to certain packages. FWIW,
> there is no way I am going to change pyepics to not install binaries of
> ca.dll and Com.dll on Windows. I'm actually kind of amazed to hear the
> objections to distributing shared libaries, but it definitely seems you and
> @dchabot are thinking mostly about Linux.


Only about Linux, of course ;)
I know that things are completely different on Windows, where applications
bringing their own copies of libraries is a common way to avoid
incompatibility trouble.
The central package managing approach on Linux is indeed orthogonal to
that: usually the distribution's packaging guidelines explicitly forbid
carrying copies of libraries, and at least strongly suggest using proper
SONAME handling. For an idea of how Linux packaging works, look at
Chocolatey for Windows.


> > > Again out of curiosity (really), why would it be a problem if a CA
> client
> > > library for language X installed its own version of libca that was
> hidden
> > > away in some package-specific directory?
> >
> > Duplication of code, inconsistent behaviour between different clients
> that
> > is hard if not impossible to debug.
>
> Do you really see this with CA clients and find that the inconsistencies
> come from different libca libraries that cannot be easily identified by
> version number? I don't think I've ever seen that.
>
>
My answer was more general.
In this case, it applies mainly to libCom, and much less to libca.


> > The client side of CA seems to move pretty slowly to me. We often have a
> > > range of libca clients at our
> > > beamlines.
> > >
> >
> > True, but libCom changes pretty drastically, and has seen e.g. inside the
> > 3.15 series quite a lot of changes and bug fixes.
>
> I'm not sure how much of libCom is actually used by libca,


Actually, a lot.
libCom provides the OS abstraction layer, so any interaction with system
services (threads, processes, sockets, semaphores, time, printing, logging
...) will use libCom. In addition, some heavily used functional modules
(queues, hashes, calc expressions, conversions) are also provided by libCom.


>
> > But, if it is a problem, would setting PYEPICS_LIBCA to force the libca
> > > used a reasonable solution?
> >
> > Maybe. Environment variables are quite volatile and not really well
> > traceable, so personally I would try to avoid it.
>
> How do you use EPICS clients that work with large data? How do you deal
> with display managers finding files? Really, I'd like to hear answers to
> those because for me, EPICS largely rests on setting environmental
> variables. I don't know how one can possibly build EPICS tools without
> setting environmental variables, and I basically cannot use EPICS without
> setting EPICS_CA_MAX_ARRAY_BYTES. So, it's kind of astonishing to hear
> that from both of you.
>

I know that setting a lot of environment variables is a very common way of
using EPICS.
However, IMHO a typical client machine should not set any EPICS related
environment variables, except for EPICS_CA_MAX_ARRAY_BYTES if needed (NB.
this will go away in 3.16).
Most other (network related) settings are often a sign of a wrong or
inadequate network setup/configuration. In complex network environments, CA
Gateways can provide the necessary proxy connections, eliminating the need
for client configuration. Setting variables unnecessarily may introduce
unwanted behavior or affect performance.

Obviously, the CA Gateways themselves (that act as client and server on
different networks) will need explicit configuration that can get quite
complex.

Building EPICS applications will need a single variable (EPICS_HOST_ARCH)
being set, but building is a separate issue and usually not done on
production machines.

Like any other application, if EPICS Base applications should be callable
from the command line without specifying their full path, the appropriate
OS mechanism (e.g. setting PATH or installation in a system location) must
be used - that's the scope of the operating system.

Display managers - or other applications - finding files (or not) is not
really in the scope of EPICS Base.


>
> What would your suggestion be for distributing pyepics in a way that could
> be used with Anaconda Python on Linux if we did not include libca/libCom
> built without readline support and did not use any environmental
> variables? I know that does not exist right now, and I don't really have a
> good idea what that would look like. OTOH, a way to distribute pyepics so
> that it can work on Linux with Anaconda or (probably) any other version of
> Python is ready to be merged.
>

Note that I never challenged your intention to distribute libca and libCom
with pyepics. In fact, I think for a plain pythonesque installation on a
system without packaged EPICS this is the appropriate way.

I did suggest that if you build a special instance (i.e. no readline) of
the EPICS Base libraries anyway, you should configure that build to provide
the libraries in the format that you need, instead of setting up a build
using SONAME and then using external tools to patch that out of the
artifacts after the build.

newville:
@dchabot -- I think I might be wearing out here.  The sort version is that
I'm not really getting the concern about using a supplied version of
libca.so by default, with the ability to override this with PYEPICS_LIBCA.

I'll give it one last try, but also apologize in advance if this seems to
have anger in it -- it really doesn't, but I am sort of finding the
discussion to be approximately endless and like we're not actually reaching
consensus.

On Wed, Apr 26, 2017 at 10:24 AM, dchabot <notifications@github.com> wrote:

> One possible runtime issue with systems that use multiple versions of the
> same library (say, libca.so.x.y and libca.so.x.z) occurs when trying to
> track down problems ("is it server-side libca.so.x.y or client-side
> libca.so.x.z?").
> But isn't that true in all cases? There is a server that uses a version of
> libca. There is a client that uses a version of libca. They aren't the
> same.
>
> That may well be true in the systems you are familiar with, but it is not
> true in the systems myself and others administer - there is one version of
> epics libs, for reasons stated previously. I don't believe it is unusual.
>

When I collect data at my 1 beamline, there are between 7 to 9 IOCs I
access, on at least 5 different computers, and 3 different operating
systems.  During the majority of the 5000 user hours per year, we access 7
or more IOCs, and most minutes, it's at least 5.  If connection to any one
of those 5 IOCs goes down, we are not collecting data.  One of these IOCs
is on the other side of the APS shield wall -- I am not sure what
generation of vxWorks that machine is running, and certainly do not control
the version of epics base that is used.  Most of the machines on our side
now use base 3.15 (and we just upgraded our main host machine, so I know
this is what most of our systems are using), but I would not be surprised
if a couple were using 3.14 -- I'm not even sure.

That's the IOCs.  For workstations using pyepics, I access PVs on those
IOCs from at least 3 computers (both Linux and Windows, and usually OSX,
though we made fade that machine out) basically full time.  For Windows,
the ca.dll I'm using is pretty old, but I think it's from 3.14.  Yeah, I'm
not sure.

I have NEVER seen a conflict due to versions of libca.   Have you EVER seen
a problem from differing versions of libca?

Do you actually see problems with the version of libca.so built with base
3.15.5 that is included with pyepics?   If so, isn't that really an EPICS
problem?  CA is supposed to be tolerant of mismatches in versions of the CA
library -- in a highly distributed system like EPICS, it's kind of vital.

Anyway, if there ever is some problem with libca or libCom that affects
pyepics (say, readline....), we can always replace it.


Another issue your proposal raises is coupling libca version to pyepics.
> This is something to be avoided, as it breaks ease of upgrading (or
> downgrading!) library versions.
> Huh? How does it break ease of upgrading library versions?
>
> With PR #80 <https://github.com/pyepics/pyepics/pull/80>, libca/libCom
> version are now coupled to pyepics version, unless the user overrides the
> now default behavior via PYEPICS_LIBCA. Upgrades to system-or-otherwise
> installed libca/libCom can't be used without resorting to setting or
> re-setting this environment variable.
>

Well, one way of looking at that is to take the 3rd party app view: Pyepics
includes libraries that are known to work with it.  Using some other
library file  named "libca.so" is fragile for pyepics, and has been shown
to cause significant problems that prevent pyepics from working.  So it's
preferable to use its own version.  That way, if your sysadmin goes and
does something crazy like build a new version of libca.so with an
incompatible version of readline, your pyepics apps will still work.


> I think PR #80 <https://github.com/pyepics/pyepics/pull/80> also
> potentially breaks packages that various groups have distributed in
> Anaconda channels by obligating use of pyepics' libca/libCom versions
> instead of any that may be packaged for use with Anaconda.
>
> What's the objection to an environmental variable?
>
> It's not the presence of the env variable that I argue against, it is the
> obligation to use that approach - I have to patch my systems. I was under
> the impression that PYEPICS_LIBCA was for scenarios where one had no other
> recourse or for testing different pyepics/libca/libCom combinations. I
> guess that was a bad assumption.
>

PYEPICS_LIBCA is available for any scenario needed.   Many people at the
APS rely on it, actually.  It's useful in cases where you have epics built
for multiple architectures and use a common filesystem for multiple
architectures / platforms.

Sorry, but the objection to environmental variables just makes no sense to
me.  Pyepics must locate libca.so to work at all. If it does not use local,
built-in libraries, it *MUST* use environmental variables (PATH,
LD_LIBRARY_PATH, EPICS_BASE, PYEPICS_LIBCA) to find that library.  We've
never used a different way.  In fact, the method proposed here is to NOT
use environmental variables, and your objecting to it... because you find
environmental variables fragile.  Huh?

Really, if you're using an EPICS client, there will be environmental
variables to set.   The complaint about having to use "fragile"
environmental variables is not a problem of pyepics.  It's a problem of
EPICS. Or maybe it's a problem with POSIX and its notion of "shell".  But
in any event, we really cannot avoid relying on environmental variables
with EPICS, especially on Unix-like systems.


> Your proposed ordering does not solve the Anaconda+readline problem. That
> is, if a system libCom.so is linked to libreadline, there is a very good
> chance that it will not work with Anaconda Python. This is the problem
> that started this topic, and it has to be solved. If we do not distribute
> and install libraries pre-compiled without readline, there needs to be
> another way to solve this problem. Do you have a different suggestion?
>
> You're right, of course. My mistake. I confused the notion of a user who
> does not have easy access to libca/libCom with the Anaconda/readline issue.
> My simplistic re-ordering of the library search algorithm won't help the
> latter.
>
> But, maybe there is another way forward? Consider this sort of
> modification to the search algorithm in epics.ca.find_libca():
>
>    1. if PYEPICS_LIBCA is set, use that
>    2.
>
> # is Anaconda python in use?if any(val in sys.version for val in ['Anaconda', 'Continuum']):
>     # is there an Anaconda packaged libca? If so, use that.
>     libca = search_conda_for_libca()
>     if libca is not None:
>         return libca
>     else:
>         # use the pyepics-provided libca
>
>
>    1. look in sys.path and the environment PATH
>    2. look under EPICS_BASE/lib/EPICS_HOST_ARCH, if those variables are
>    in the environment
>    3. fallback on the pyepics-provided libca, as no others were found
>
> The result could be confusing in mixed systems as to which libca version
> pyepics is bound to, but I don't have an immediate solution for that.
>

If we have a version of libca that works, why should we search for
anything?  There is a much higher chance of finding an non-working version.

What is "Anaconda package libca"?  Why does such a thing even exist?   Does
such a thing exist?  I don't see it on Anaconda.org.   Who maintains this?
Why is this better than including those libraries in pyepics?

If a user or admin knows they want to override the built-in libraries,
they would not look at this list of rules and say "ah, just make sure
EPICS_BASE and EPICS_HOST_ARCH are what I want " (what if the one they want
doesn't match that string?) or "ah, just put it in PATH" (because, who puts
libraries in PATH?).   They would use PYEPICS_LIBCA because that is the
first thing in the list, and it exactly sets the absolute library path to
be used without affecting any other applications.   If a shell script was
going to launch a pyepics app, it could easily export this just before
running.   It *is* the preferred approach.


Sorry, but I kind of think we need to move on.

simongregorebner:
I would like to express my full support for the view of @newville in his last post.
Packaging libca/libcom with pyepics and having it the default to be used except explicitly opting out of this is the only "clean" way also in my view!

Some people just see pyepics as an add-on to the epics base codebase, but it isn't. It is a self contained package (or at least it should be in my view) with a contract to the outer world that is nothing more than the Channel Access protocol and the Python interpreter.
Pyepics is a Python library for Channel Access, it lives in the Python "world", the Python interpreter is its base, not the operating system. Therefore the view of mixing operating system packaging with python packaging (by default) seems to be very odd to me.

If I install pyepics via a Python packaging mechanism, what I, as an application developer,  expect from it is to work out of the box. I don't want to have to install this and that operating system package, and I don't want to care what else is installed on operating system level.
The only way for pyepics to provide this is to bring the required libs with it and to use these libs by default.
For example, I guess, most people in the Python world would expect from  numpy to behave exactly the same way regardless of what operating system packages are installed (except actively tweaking things). Also these people would expect that they are not need to install any c libraries when they install numpy with the e.g. the pip command, right?

As I said before, pyepics is a Python library for Channel Access, that it uses the C-Implementation for doing Channel Access is just a matter of fact. Theoretically It could equally good use a pure Python implementation for doing Channel Access.  That people are not seriously working on this is just a matter of time and because of the unbelievable fact that the Channel Access protocol is not formally specified anywhere except in the C code (thats an extremely embarrassing fact of the  Epics collaboration the core developers should really care about).
In the same direction, how do you regard all the Java applications that use the pure Java implementation of Channel Access? Are they not debug-able/stable/etc.?

Also I would like to challenge the point of that all the libraries and applications on a system should only use the system provided libs (by default) in a different way. This kind of thinking, lets say, is a very traditional Linux view in my view. Why do nowadays most applications even bring their own OS with it - i.e. with Docker? Stability, testability, reproducibility, ... ?

Finally, having these random thoughts said, I would like to again express my full support of the idea to package the libs directly and to use them by default. I believe that this is the most user friendly, maintainable and "pythonic" way.

ralphlange:
I completely agree, however, my use case is a bit different:
- Server  (IOC) side: ~300 servers, ~1500 IOCs, ~5 million EPICS records
- Client (Control Room) side: ~30 servers, ~70 operator workstations (3 screens each)
- Full traceability of everything required.
- No development within that system.
- No transfer of binaries from the outside.

However pythonesk and appropriate your approach is for your application, it will simply not work here.

simongregorebner:
So if you agree, great!

Based on your statements the ITER system seems to be pretty easy to handle anyway:
> Full traceability of everything required.
> No development within that system.
> No transfer of binaries from the outside.

So, based on these statements, you have to fix the pyepics version anyway, i.e. nobody than the systems responsible group can upgrade, etc.. So then simply choose your version, opt out of the default by setting PYEPICS_LIBCA and there you go. You do this one time, and everybody on your site is ready to go.
Its more tricky in a dynamic environment, which is more often the case I guess, where people bring their own machines and need to quickly "connect" them to the other infrastructure. To make things stable there, as stated in the thread before, the "straight forward" way to get to a stable and predictable system is to do as @newville suggested - pyepics bings the binaries with it and uses these by default. If people pretend to know what they do, they have the option to actively opt out of this.

ralphlange:
Thanks for your very valuable instructions.
Unsubscribing.

simongregorebner:
It was not my intention to screw up people with my comments, nor to give anybody instructions on how to do things! My intention was to express how I see things as I had this very same discussion many times before.
In case I phrased things in an offensive way I am really sorry about that!

newville:
@simongregorebner  Thanks. That's nicely said, and  I agree completely with your view of how pyepics (tries to) bridge the Python and Epics world.  Aiming to be a simple way to get to EPICS PVs from within Python is definitely how I think about the problem too.  And there is definitely a move toward per-user installs of Python environments (like Anaconda, but others too) into home directories of otherwise unprivileged users. 

My view is that most pyepics users (and the ones we need to make things easy for) are not the "just recompile epics base without readline" sort of people.   The people who subscribe to Tech-Talk or are setting up systems with multiple IOCs can probably handle a bit of Python or shell scripting to override the basic defaults.  

@dchabot @ralphlange If there is a concern about using any outside binary files (probably based on perfectly sensible security concerns),  would it be reasonable to add an option to "setup.py" to *not* install the shared libraries?   That would allow the search of PATH/EPICS_BASE etc to find the system version, provided a local version was not installed.   That would require installing from a command-line (perhaps `python setup.py --no-shared-libs install` or similar), but maybe that's actually preferred for admins of larger systems or systems with stringent security requirements.

Like @simongregorebner says, really there is absolutely no ill-will or anger here, just honest differences in perspective.

danielballan:
You outed us! It's true. Caproto has accidentally evolved from a side-project to a somewhat serious endeavor, but we still have a bit of work to do before we announce this on tech talk. Drop by if you want to help. Very interested in everyone's thoughts on this approach, but give us a bit of time to clean up before you dig too deeply. :- )

newville:
@dchabot   OK, I'll pull the changes here to master.  Adding an option to `setup.py` of "no-shared-libs" or similar would be fine with me.  If you're up for that, great!

newville:
@dchabot @simongregorebner @danielballan  One of the original design criteria for pyepics (back when ctypes was new and cool and there were several competing wrapping of CA, all at least slightly incompatible) was that an explicit and complete wrapping of libca was *better* than trying to rewrite or reverse engineer the CA protocol.   That is all the hard work that went into libca and libCom is readily, directly exposed to Python via ctypes.  This gives both a very useful reference and many man-years of debugging.  So pyepics'  `ca` and `dbr` modules started quite deliberately with as-close-as-reasonable mappings of the C implementation, and then built from there.

So: why rewrite the CA protocol in Python?  Is this related to Epics V4?

FWIW I believe that Java had/has both a re-implementation of CA and a wrapping of the C implementation (via jni).  I don't pay much attention to Java, so I'm not sure if both are in common usage.

danielballan:
The _original_ personal motivation was: I was stuck inside my apartment for a week recovering from surgery and I wanted to understand Channel Access better and explore the growing world of concurrency libraries in Python.

The emerging justification for taking this more seriously is the ability to pull apart network transport, message parsing/composition, and message processing. See this paragraph: https://danielballan.github.io/caproto/doc/#why-do-this

newville:
@danielballan  OK. I'm not sure I really get it, but I'll have to look into this in more detail.

simongregorebner:
@dchabot @danielballan These are very interesting news, thanks a lot for sharing! I definitely will have a look at it. In case you need some ideas on how to write certain pieces of channel access and you know how to read Java, here the code to "our" complete rewrite of the Java channel access library: 
https://github.com/channelaccess/ca

danielballan:
@simongregorebner Thanks Simon. I discovered your API documentation in the first week of working on this, and it was helpful in the early stages of trying to grok the different commands' meanings and quirks.

newville:
@danielballan   I've only had time to read the basic motivation. It might be my lack of imagination, but I'm not sure I understand the idea of sans-I/O argument.  Is the idea to be able to use asyncio and coroutines with a direct socket layer?  That could be very interesting. Is this also intended to provide a CA server?  

Also out of curiosity: if this is "better" (for some definition) would the idea be to eventually migrate this back to C to make it available for more widespread use?  Because, one of the appeals of using libca is that it uses exactly the same interface as all other (except Java?) CA components.  For reference, the discussion earlier in this thread included concerns about pyepics using the different *versions* (point releases even) of libca from the IOCs.   That is, some folks were not thrilled about using a pre-compiled version of libca packaged and installed with pyepics, but wanted to make sure pyepics would use the libca compiled on their system.

danielballan:
Thanks for taking the time, @newville.

> I've only had time to read the basic motivation. It might be my lack of imagination, but I'm not sure I understand the idea of sans-I/O argument.

To start, I spent a lot of time absorbing h11 (sans-I/O for HTTP 1.1). Nathaniel probably explains the motivation better than I do.

> Is the idea to be able to use asyncio and coroutines with a direct socket layer? That could be very interesting.

Yes, exactly.

> Is this also intended to provide a CA server?

Yes. We have a simple working server example, and it's pleasantly simple to set up. The design of caproto (imitating the design of h11) makes clients and servers as symmetric as possible, so servers use mostly the same code paths as clients.

> If this is "better" (for some definition) would the idea be to eventually migrate this back to C to make it available for more widespread use?

I guess "migrate back to C" would mean refactoring libca to take a sans I/O approach: to keep network calls in separate functions from the parsing and state-updating code. It is possible in principle if people decide they want to do it.

PR #83, opened 2017-05-01
Fix libca install optout

tacaswell:
attn @dchabot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,8 +61,10 @@ install:
   - conda install numpy
   - pip install coveralls pytest pytest-cov
   # install pyepics as a source install
+  - ls -l /home/travis/mc/envs/testenv/lib/
   - pip install -v .
-
+  - ls -l /home/travis/mc/envs/testenv/lib/
+  
   # setup some path environment variables for epics
   - export PATH=$PATH:$EPICS_BASE/bin/$EPICS_HOST_ARCH
   - export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$EPICS_BASE/lib/$EPICS_HOST_ARCH"
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ before_install:
   - conda config --add channels soft-matter
 
   # MAKE THE CONDA RECIPE
-  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION
+  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION epics-base
   - source activate $CONDA_ENV
 
   # need to reactivate after installing epics-base so that the EPICS_BASE env
@@ -62,6 +62,8 @@ install:
   - pip install coveralls pytest pytest-cov
   # install pyepics as a source install
   - pip install -v .
+  
+  - find /home/travis/mc -name 'libca*'
   
   # setup some path environment variables for epics
   - export PATH=$PATH:$EPICS_BASE/bin/$EPICS_HOST_ARCH
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ before_install:
   - conda config --add channels soft-matter
 
   # MAKE THE CONDA RECIPE
-  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION epics-base
+  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION
   - source activate $CONDA_ENV
 
   # need to reactivate after installing epics-base so that the EPICS_BASE env
@@ -61,9 +61,7 @@ install:
   - conda install numpy
   - pip install coveralls pytest pytest-cov
   # install pyepics as a source install
-  - ls -l /home/travis/mc/envs/testenv/lib/
-  - python setup.py install
-  - ls -l /home/travis/mc/envs/testenv/lib/
+  - pip install -v .
   
   # setup some path environment variables for epics
   - export PATH=$PATH:$EPICS_BASE/bin/$EPICS_HOST_ARCH
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ before_install:
   - conda config --set always_yes true
   - conda update conda --yes
   - conda install conda-build anaconda-client jinja2
+  - conda config --add channels lightsource2-tag
   - conda config --add channels lightsource2
   - conda config --add channels soft-matter
 
@@ -60,7 +61,7 @@ install:
   - conda install numpy
   - pip install coveralls pytest pytest-cov
   # install pyepics as a source install
-  - python setup.py develop
+  - NOLIBCA=1 pip install -v .
 
   # setup some path environment variables for epics
   - export PATH=$PATH:$EPICS_BASE/bin/$EPICS_HOST_ARCH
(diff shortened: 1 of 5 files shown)


newville:
@tacaswell  Thanks!   The approach of using `NOLIBCA=1` at install time seems fine to me. 

I'd also like to give @dchabot,  @ralphlange, and @simongregorebner if this approach is acceptable to them.

.travis.yml
```diff
@@ -60,7 +61,7 @@ install:
   - conda install numpy
   - pip install coveralls pytest pytest-cov
   # install pyepics as a source install
-  - python setup.py develop
+  - NOLIBCA=1 pip install -v .
```
dchabot: Doesn't this need an `export`?

INSTALL
```diff
@@ -1,39 +1,24 @@
...
+By default the object libraries for Epics Channel Access (libCom.so
+and libca.so for linux, libCom.dylib and libca.dylib for Mac OS X and
+Com.dll and ca.dll for windows) are included with this distribution
+and installed for you.  To prevent the installation of the binaries use
 
-This code has been tested most extensively with Python2.6 on 32-bit Linux
-and Win32 systems, and versions of Epics base 3.14.10 or 3.14.11.  The
-codes are also tested (though admittedly less regularly)  with other
-combinations of Platform, Python, and Epics base, including:
+   NOLIBCA=1 python setup.py install
```
dchabot: `export NOLIBCA=1` ?

.travis.yml
dchabot: The libca version used by Travis is different between py2.7 and py3.6 (didn't chk other py3.x versions:

py2.7:
```python
$ python -c "import epics.ca; print(epics.ca.find_libca())"
/home/travis/mc/envs/testenv/lib/libca.so
```
py3.6
```python
$ python -c "import epics.ca; print(epics.ca.find_libca())"
libca.so.3.14
```

tacaswell:
I spent some time trying sort out how to use setuptools and decided to use the sledge hammer instead ;)  I will not be offended if you want to re-do this properly.

My understanding of environmental variables was that the `NOLIBCA` would be there in the environment that `python` executes in, but would not pollute the global namespace.

Only 3.6 is different, which is odd.  The files installed by the epics-base conda package should be at /home/travis/mc/envs/testenv/lib/ and it looks like the `*.so` files are links to the `*.so.3.14` files (see https://anaconda.org/lightsource2/epics-base/files), but I can not reproduce that locally....

newville:
@tacaswell  I thought using an environmental variable was a nice approach.  

I probably would have done a simple check for `--nolibca` in `sys.argv`, acting if it was set, and removing it from `sys.argv`.

No guesses on differences between 3.5 and 3.6.  More `setuptools` magic?

simongregorebner:
The variable approach looks good to me (it seems that also numpy is using this way to enable/disable stuff - see https://docs.scipy.org/doc/numpy-1.10.1/user/install.html#disabling-atlas-and-other-accelerated-libraries).

I checked with the official miniconda3 docker image from continuumio (i.e. Python 3.6) and for me this approach also seems to work cleanly for 3.6 (the /data folder I mount is the clone of the pull request, the pip version within this docker image is 9.0.1)

Here the truncated output of what I didd
```
$ docker run --rm  -v `pwd`:/data -it continuumio/miniconda3
root@7d31fd699bec:/# cd /data
root@7d31fd699bec:/data# NOLIBCA=1 pip install -v .
Processing /data
[... lots of output ...]
done
  Removing source in /tmp/pip-of5yv_aq-build
Successfully installed pyepics-3.2.5rc3+143.g0c64e92
Cleaning up...
Looking up "https://pypi.python.org/pypi/pip/json" in the cache
No cache entry available
Starting new HTTPS connection (1): pypi.python.org
"GET /pypi/pip/json HTTP/1.1" 200 72983
Updating cache with response from "https://pypi.python.org/pypi/pip/json"
Caching b/c date exists and max-age > 0

root@7d31fd699bec:/data# python --version
Python 3.6.0 :: Continuum Analytics, Inc.

root@7d31fd699bec:/data# python -c "import epics.ca; print(epics.ca.find_libca())"
  File "<string>", line 1, in <module>
  File "/data/epics/ca.py", line 194, in find_libca
    raise ChannelAccessException('cannot find Epics CA DLL')
epics.ca.ChannelAccessException: cannot find Epics CA DLL

root@7d31fd699bec:/data# pip uninstall pyepics
Uninstalling pyepics-3.2.5rc3+143.g0c64e92:
[... lots of output ...]
Proceed (y/n)? y
  Successfully uninstalled pyepics-3.2.5rc3+143.g0c64e92

root@7d31fd699bec:/data# pip install -v .
Processing /data
  Running setup.py (path:/tmp/pip-6yn3ra03-build/setup.py) egg_info for package from file:///data
    Running command python setup.py egg_info
[... lots of output ...]
done
  Removing source in /tmp/pip-6yn3ra03-build
Successfully installed pyepics-3.2.5rc3+143.g0c64e92
Cleaning up...

root@7d31fd699bec:/data# python -c "import epics.ca; print(epics.ca.find_libca())"
/opt/conda/lib/libca.so

root@7d31fd699bec:/data# conda list pip
# packages in environment at /opt/conda:
#
pip                       9.0.1                    py36_1  
```

newville:
@tacaswell @dchabot I'm not sure I am completely understand the Py 3.6 vs 3.5 issue.  Like, is this a real problem that should be solved before merging?    

Feel free to merge this when it's working for you all...

newville:
@dchabot I think that #82 set the RPATH to be $ORIGIN for both `libca.so` and `libCom.so`.   Or, at least, that was the intention.  I don't quite understand when / why that would fail.

Trying to mess with `setuptools` seems like real work to me.  I would probably have not tried anything that fancy, but just done a test for `--nolibca` in `sys.argv`.

newville:
@dchabot 
> Meh, it's a learning experience. Gives me something to do during long meetings that I probably shouldn't be invited to ;-)

OK, glad we could help distract you during ~~boring~~ important meetings!

tacaswell:
I am super confused why https://travis-ci.org/pyepics/pyepics/jobs/226602598 worked....

Reverted the one small logic change and tweaked the test matrix to try both with and without NOLIBCA

@dchabot GH should give you rights to push to my branch, feel free to do so!

tacaswell:
ah, that makes sense.  Could change it to `pip install -ve`, but that seems like cheating.

Is the issue here that while we are over-riding the normal search path for libca, but letting the system take care of it's self for libcom?

I see that not installing epics-base from conda makes some things fails (because the binaries are missing).  We could use https://github.com/NSLS-II/ophyd/blob/master/install_epics_on_travis.sh and install a system-level epics instead of the conda-level one.

tacaswell:
To be honest I am a bit surprised that that worked....but travis is happy now!

newville:
@tacaswell @dchabot Could that be related to #82 and the setting of RPATH for the dlls created in the NSLS-II epics base image?  That is, would it work to do
```
patchelf --set-soname libca.so  libca.so
patchelf --set-soname libCom.so  libCom.so
patchelf --set-rpath '$ORIGIN' libca.so
patchelf --set-rpath '$ORIGIN' libCom.so
patchelf --replace-needed libComs.so.3.Whatever libCom.so libca.so
```

I think Ralph Lange suggested a way to do that at epics-base build time too....

Anyway, if this is ready to merge, go for it.

PR #85, opened 2017-06-15
Iron python testing

hartmansm:
In follow up to http://www.aps.anl.gov/epics/tech-talk/2017/msg00797.php . . .

With this patch (based on work by @FreddieAkeroyd), pyepics is now working with IronPython (32-bit only). Tested on Windows 7 with IronPython 2.7.7 (2.7.7.0) on .NET 4.0.30319.42000 (32-bit)
using ca and Com dlls included with the pyepics distribution. 

Patched from tag 3.2.6.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -239,7 +239,11 @@ def initialize_libca():
 
     # save value offests used for unpacking
     # TIME and CTRL data as an array in dbr module
-    dbr.value_offset = (39*ctypes.c_short).in_dll(libca,'dbr_value_offset')
+
+    # in_dll is not available for arrays in IronPython, so use a reference to the first element
+    value_offset0 = ctypes.c_short.in_dll(libca,'dbr_value_offset')
+    dbr.value_offset = ctypes.cast(ctypes.addressof(value_offset0), (39*ctypes.c_short))
+
     initial_context = current_context()
     if AUTO_CLEANUP:
         atexit.register(finalize_libca)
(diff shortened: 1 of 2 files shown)


newville:
@hartmansm @FreddieAkeroyd -- this looks great!  Thanks so much!  Most of this appears to be well-isolated to IronPython, but I'd like the chance to evaluate and test this branch with CPython on Windows and other platforms.   

But I also think that barring any unforeseen issues, we should plan on merging this soon.  Unfotrunately, I'm doing a lot of travel this and next week, so it may not be until early July until I have a chance to evaluate this fully myself.

Does anyone have objections to this PR or to waiting a few weeks to allow thorough testing?

epics/ca.py
```diff
@@ -584,7 +591,10 @@ def _onGetEvent(args, **kws):
     if args.status != dbr.ECA_NORMAL:
         return
 
-    get_cache(name(args.chid))[args.usr] = memcopy(dbr.cast_args(args))
+    if dbr.isIronPython():
```
klauer: This function call will be done on every `get` callback from libca. It should be just made into a boolean in `dbr.py`.

hartmansm:
Thanks, @newville. There is no urgency for the merge. This is for one application on one beam line on a non-centrally managed machine for vendor software. They can install from the branch if needed.

@klauer, I just pushed a commit to replace the function call.

Issue #90, opened 2017-09-06
pyepics PIP package is installing unwanted shared libraries


hdante:
The following files were installed with pyepics, but they shouldn't have been:

  /usr/lib/libCom.so
  /usr/lib/libca.so
 
 Instead, pyepics should use the system provided EPICS libraries.

 Extra: these libraries caused problems when installing other unrelated packages with yum. Example follows attached.

[yum-error-log.txt](https://github.com/pyepics/pyepics/files/1282448/yum-error-log.txt)

newville:
@hdante  Sorry for the trouble.  Yes, with 3.2.7 epics now installs libca and libCom on Linux, Mac, and Windows.  By default. these get installed according to the install method. For the pip install you did, they go to /usr/lib (which makes sense for "/usr/bin/python").   That's how pip install with a typical system-supplied python will work.

Sadly, we decided earlier this year that using "system provided EPICS libraries" was causing too many unresolvable problem.  The default build of libCom.so has the very nasty behavior of linking against libreadline.  Of course, CA does not use readline.  Having libca.so depend on libreadline.so leads to unresolvable conflicts for applications that want to use pyepics but have also linked against a different version of readline (maybe those applications actually use readline).  This conflict does really happen for many folks using a non-system supplied version of Python, such as Anaconda Python.

As a result, we've decided to build and distribute versions of libca.so and libCom.so that are not linked with readline, hoping that these would be suitable defaults for most users of pyepics.   In short the real change was from defaulting to "system-supplied libca" to "not-contaminated-by-readline libca".  You can always specify which version of lib CA to use by setting PYEPICS_LIBCA. 

It seems like the main problem you are seeing is that these versions of libca.so and libCom.so do not work for you, though I don't know why that is.  Is this RHEL6.9?   We've tested the version we distribute on a few different linuxes (including using a docker image with TravisCI), but maybe there is a conflict with RHEL6.9....   I also don't know why yum would complain about these.  Perhaps the files are invalid on RHEL6.9?    Anyone have any thoughts on why that might be?

hdante:
I'm not sure what the messages meant, but it would be reasonable for it not to go into a standard path. It could go to $libdir/pyepics, instead, for example, so that they're isolated only for pyepics use. This is how other libraries store their internal shared libraries. Examples follow:
[henrique.almeida@sol13-linux lib64]$ ls -lR /usr/lib64/openssl/
/usr/lib64/openssl/:
total 4
[...]

Issue #91, opened 2017-09-12
access_rights callbacks on Python2 Windows64


newville:
callbacks for access_rights on Python2 + Window64 are wrong.  I think this is probable due to problems similar to #31 and probably requires a special case at  [epics/dbr.py#L348](https://github.com/pyepics/pyepics/blob/70b372e522ded44848a23dc88075740f67861023/epics/dbr.py#L348)

newville:
merged.

Issue #89, opened 2017-09-06
Exception thrown on python exit


hdante:
Red Hat 6.9, EPICS base 3.14.12.5, python 3.4.5, pyepics 3.2.7

[henrique.almeida@sol13-linux ~]$ python3
Python 3.4.5 (default, Jun  1 2017, 13:52:39) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-18)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import epics
>>> epics.caget('LNLS:ANEL:corrente')
177.617
>>> 
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/epics/ca.py", line 251, in initialize_libca
  File "/usr/lib/python3.4/site-packages/epics/ca.py", line 213, in find_libCom
  File "/usr/lib/python3.4/site-packages/epics/ca.py", line 178, in _find_lib
  File "/usr/lib64/python3.4/ctypes/util.py", line 233, in find_library
  File "/usr/lib64/python3.4/ctypes/util.py", line 203, in _findSoname_ldconfig
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2222, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 2155, in _find_spec
TypeError: 'NoneType' object is not iterable

newville:
@hdante Hm, I don't see that on Centos7 / python 3.6, but I don't have a RHEL 6.9 machine to test on.

Am I reading that correctly that this happens within the interactive REPL?  Or is it happening when exiting a very small script?

I *have seen* problems that might be similar with some very scripts that create a single PV and immediately exist.  It sort of seemed like "initialize_libca()" was being run twice, as if it was running at exit.

Not that this is an actual solution, but:  Does the error go away if you put a short `time.sleep` before exiting?

hdante:
@newville This happens in interactive shell, in a small script and in a small script with sleep(1) as the last statement.

I'm sure initialize_libca() is running at exit, after the last statement. The problem never happens before it.

hdante:
Hint: the remaining hidden backtrace goes back to line 832 at pv.py, on PV.disconnect(), called by PV.__del__().

hdante:
[...]

Issue #88, opened 2017-09-02
why does autosave uses PV() instead of pv_get()?


mennodoor:
Hi,

I am really not sure about this, but wouldn't it be saver to use pv_get() instead of PV() to prevent creating lots and lots of objects when this script is used for a data logging system?

regards
menno

newville:
@mennodoor You are right.  Using PV() instead of get_pv() to create objects inside the `autosave` functions will create a PV object for the same Process Variable name each time the function is called.  The underlying connection will be reused, but it definitely would be better to use `get_pv()`.   A PR is welcome!  

Actually, looking at `autosave.save_restore.py` more closely, it might be worth considering changing this to create an AutoSaver class that parsed a request file once and had a `save()` method.  That might be better for the use case of data logging, as the PV objects could be created and connected once and then saved to the proper format more easily, perhaps controlling the naming of sequences of saved files as well.

But, sadly, I *just* released 3.2.7 yesterday (3.2.6 was more than a year earlier).  There are a few outstanding issues, but I think most of them are in on of the categories of "actually resolved, but not confirmed", or "not resolved but apparently rare, and not a clear solution", or "reminders of work that could be done".  I'll ping these outstanding issues over the next week, but would be glad to hear any other suggestions for improvements.

newville:
now fixed to use `get_pv()` in #98, which also adds an `AutoSaver` class for repeatedly saving save files from a single request file.

newville:
fixed in master, soon to be released in 3.3.0

PR #92, opened 2017-09-15
MNT: c library installation 

dchabot:
Addresses #90.

This PR will result in c library installation to "\<install-location\>/epics/clibs/\<os-arch\>/\<libname\>".
Works similarly for `python setup.py install` and `pip`.

Users can still choose to forgo 'libca/libCom' altogether by:

```shell
> NOLIBCA=1 pip install .
```


newville:
@dchabot  Thanks very much -- that looks really good to me (without having tried it yet). 

A few questions that perhaps you know the answers off the top of your head, but I don't:

1.  Is it possible to include all of `epics.clibs` for all installations?  Is it desirable?   I don't have a strong preference, but it might be simpler if we always install exactly the same files.

2. Is it possible to test "pip install" from a local source kit separately from "python setup.py install"?   It must be, but I've never looked into how to do that.

3. Should we remove "Test 4" (search through "known, typical" install locations) from `ca._find_libca()`.  This legacy search should probably be dropped at some point.

I would like to get these figured out and merge this and also address #88 and #91 for a release of "3.3.0", hopefully within a few weeks.   Does that seem reasonable to you?

FWIW,  I do not have any idea what to do about #89, and there are still some issues (#68 for example) that I think do not have a good solution.

newville:
@dchabot 
>> Is it possible to include all of epics.clibs for all installations? Is it desirable? I don't have a strong 
>> preference, but it might be simpler if we always install exactly the same files.
> It is possible, but IMO not desirable.

OK. I'm completely fine either way.

>> Is it possible to test "pip install" from a local source kit separately from "python setup.py install"? 
>> It must be, but I've never looked into how to do that.
> Do you mean have a "system" installation and a development installation coexist? That seems possible,
> but whichever module is parsed first in the sys.path will win on import, of course.

I'm wondering if it is possible to do a "pip install" from a non-PyPI source for testing.  And, yes, I now remember that you can specify "--trusted-host".

The point is that "python setup.py install" and "pip install"  gave very different results, including the basic complaint from #90 that "pip install" installed the libraries to `/usr/lib`, whereas "python setup.py install" does not.  I don't know that we can automate that test, but I'd like to try it before pushing to PyPI.

>> Should we remove "Test 4" (search through "known, typical" install locations) from ca._find_libca(). 
>> This legacy search should probably be dropped at some point.
> Yes. I agree with removing this from the library search algorithm. I can add such a commit 
> to this PR if you would like.

Yes, that would be great.  Please do.

> Another thing that strikes me as somewhat misleading is the stdout reporting of libca usage 
> upon installation:
> 
> Installing collected packages: pyepics
>  Running setup.py install for pyepics
>    UPDATING build/lib.linux-x86_64-2.7/epics/_version.py
>    set build/lib.linux-x86_64-2.7/epics/_version.py to '3.2.6+102.g6ab9928'
>
>      Will use CA library at:  epics/clibs/linux64/libca.so
>
> My problem with this is that the path reported is likely to change between installation and usage.
> Do you think it is enough to simply report if a libca could not be found?

I'd be OK with just dropping this message.  It is often Just Plain Wrong. 

It might be nice to somehow get a reported of the complete path of the CA library that will really be used.  This isn't really available right now, and might be somewhat complicated, as `ctypes.util.find_library()` does not give a full path. Hmm, maybe it's not that easy...

> I have no idea about #89, as I've not seen that before. I have seen odd shutdown behavior locally 
> when using pyepics and the pytest library. Namely, a failure to exit. It will not respond to ctrl-c and 
> must be kill -9'd from another terminal. Light investigation indicated flow control was stuck in 
> epics.ca.finalize_libca(). Have not looked further, but this may not be related at all.

Yeah, I still sort of suspect that the hint I gave about `ca._find_libca()` in the discussion of #89 might be relevant (as I said, I did find one such problem when doing final testing of 3.2.7rc1 on Windows).  But this sort of problem is very hard to debug without actually seeing the problem.

newville:
@dchabot  Hm, yeah the results from Travis sort of confuse me too.

Do we still need the NOLIBCA install-time option?  Maybe it's better to always install (our best guess at a working version of) libca to epics/clibs, and allow PYEPICS_LIBCA to override using that at run time?   If nothing else, it would allow us to be confident that a version of libca will always be installed, and also know its path relative to ca.py.

For testing:  You are right that the Travis tests appear to happen within the cloned 'epics' directory.  That's probably a bad idea.   Adding 'cd tests' as the first command of the "script" paragraph of 
.travis.yml should ensure that the *installed* version of epics is used rather than the local version.

newville:
@dchabot 
> I can work on the CI tests tomorrow. They could stand for some general cleanup as well (I 
> see some debugging info in there from a previous PR).

Great! The Travis-CI config does look a bit messy to me: I don't think we need all the non-standard channels, the coverage reports, running gdb on py.test failure, etc.

> IMO, an install-time option to forgo installation of packaged libca/Com is a Good Thing, 
> for reasons argued in other PRs/issues.

My perspective is that if libca is always installed into a non-harmful, easily findable location, then that can always be available, even if only used for testing.   When someone comes with a question or problem, we'll know whether that file was installed.

How would a user know to install with NOLIBCA?   It turned out to not work for #90, and appeared to not be a desirable workaround for those users. One might have assumed had some experience with both Epics and Python, which might have made them somewhat blind to the obvious.

Pyepics is client software, definitely aimed at people who may not know anything else about Epics (like "what's Epics Base?").  That user wants to do some CA with some version of Python on the OS of their choice.  They do not need epics base or any other epics tools installed.  They only need libca and pyepics.   We should make sure pyepics works for them.  The folks developing IOCs and writing Epics drivers will probably be able to figure it out, eventually.

newville:
@dchabot  Isn't the idea here to put libca into something like `sys.prefix`/site-packages/epics/clibs so that it cannot conflict with other libraries?   In that case, what is the potential for harm in always installing to that pyepics-only location? 

I don't think it's a big deal to keep the NOLIBCA install option. I doubt it will be used much, which suggests to me it will be of little value.  I don't know who it will really help. 

> Including the c libraries is a convenient means to help that category of user get up and running 
> quickly. But in doing so, it introduces a bug that can interfere with systems where the c libraries 
> are installed by other methods.

I thought this PR would avoid that conflict. (For the record, I do sees this as a conflict not a bug.  It is very specific to Linux and installs to system python.  But we should try to avoid the conflict to the extent that we can.) 

> To assume that pyepics is the only EPICS software present on a system is neither realistic 
> nor sound defensive engineering.

I agree. I don't think we are making that assumption. 

I think we concluded several months ago that pyepics needs to come with working (to the best of our abilities) versions of libca on all systems.  I still think that is true.  My preference would be to have a simple and unifom installation process.

Having pyepics install libca.so to /usr/lib has undesirable consequences in that it apparently can conflict with /usr/lib/libca.so installed by other methods.   But having pyepics install libca.so to `sys.prefix`/[site,dist]-packages/[py]epics[.xxxxegg]/clibs should be OK, no?   

I never really understood the complaints from yum about /usr/lib/libca.so.   I don't know what they mean.

newville:
@dchabot 

> If we always install the pyepics-packaged C libraries, then there are no simple means to have
> pyepics use system-provided C libraries: the search algorithm in find_libca() would always 
> chose the former over the latter.

Use PYEPICS_LIBCA?   

> And sorry, but I don't consider using the run-time, PYEPICS_LIBCA, a good solution to what
> is an installation-time problem.

Really? Isn't finding a shared library a runtime problem, kind of by design?  If libca.so changes after installation or you somehow point to a different version, you will get different behavior.  This is generally thought of as a feature, but sometimes as a nightmare.  It's kind of the way modern OSes work.  You *never* set a shared library at install time.  Well, I guess unless you install a private version and only look where you installed that too.  That's kind of what we're doing here, but we're providing a way out for those that think they know what they're doing.

For non-python apps, one would expect the location of libca to be found by configuring ldconfig or using LD_LIBRARY_PATH.  Even on Linux, Epics base does not normally install libca.so to "normal linux lib directories" like /usr/lib or /usr/local/lib (at least not on any platform I've ever seen), so you sort of have to set these up.  It's kind of a pain, actually, but it seems kind of unavoidable.

Anyway, why not use a dedicated env variable that explicitly says where pyepics should find libca?

> Having the option to not use the pyepics-packaged libraries will help me and the facility I work at.

Um, sorry but I just do not understand this.  Why is that?    

Honestly, I really do not understand why anyone would want to use something other than the supplied libca.so --the python code probably won't be able to use new features anyway.  If there are bugs or problems in libca.so that affect pyepics, we should fix them.

What do you want to do?

newville:
@dchabot .. but also:  keeping the NOLIBCA install-time option is OK with me, I just thought it wouldn't be necessary with the PR.

newville:
> Honestly, I really do not understand why anyone would want to use
> something other than the supplied libca.so
> Epics base does not normally install libca.so to "normal linux lib
> directories" like /usr/lib or /usr/local/lib (at least not on any platform
> I've ever seen), so you sort of have to set these up.
>
> Is the notion of "system integration" perhaps being overlooked?
>

Well, I don't think so, but maybe I'm thinking about that differently from
you.

Removing NOLIBCA would omit one avenue for achieving that. Encapsulating
> the pyepics-packaged libs (as this PR seeks to do) to prevent "conflicts"
> is a related system integration matter.
>

If the libs are put into only-found-by-python directories, doesn't that
achieve the same result?  It seems to me that one would want NOLIBCA to
*avoid* system integration concepts, and use some *untested* version of
libca.so instead of the standard one.  Maybe that's a different perspective.

If pyepics is installed via system packaging (yum, apt), the libraries end
> up in /usr/lib. We use Debian packaging at my facility (so does NSLS2 and
> others, although NSLS2 beamlines rely on Anaconda too), and it sounds like
> the op in #90 <https://github.com/pyepics/pyepics/issues/90> is using
> rpms to achieve the same thing. In these cases system integration is an
> important facet to consider, and ctypes.util.find_library() simply does
> The Right Thing to locate libca.
>
> I hope that makes sense.
>

Are there existing yum or apt packages for pyepics?  I wasn't aware that
there were, and certainly not aware of anyone using them.  Anyway,
ahouldn't those be configured to also install libca.so specifically for
pyepics within the python install (epics/clibs)?

The op in #90 did "[sudo] /usr/bin/pip install pyepics".  It worked the way
you describe: installing libca.so to /usr/lib/libca.so.   That was exactly
the complaint that we are trying to address.

Again (sorry for repetition, but it does appear my point is not being
understood),  That install to /usr/lib/libca.so happens ONLY with
`/usr/bin/pip install pyepics`.  It does not happen with `/usr/bin/python
setup.py install`.  That complaint could certainly be addressed with any of

1. don't install to system python.
2, don't install with pip (perhaps using a package manager instead?)
3. install with  NOLIBCA=1 pip install pyepics.
4. install with a source kit and /usr/bin/python setup.py install

None of those options appeared to be sufficient.   So, we want to make sure
(ie test) that libca.so can not be installed to /usr/lib/, even when using
`pip install`.

Again, the NOLIBCA option is OK, it just seems to me like it *breaks*
system integration in that it will use a version of libca.so that is not
tested with pyepics, will add an unclear and fragile dependency of pyepics
on some un-specified version of libca.so, and will make it slightly harder
for us to troubleshoot problems.  What makes it acceptable to me is that it
is very likely that no one will use it, and that if someone does use it and
runs into a problem with libca.so, we can point to an easy solution of
reinstalling with libca.so.

stuartcampbell:
@dchabot Just to chime in, and give a :+1: for keeping the NOLIBCA option.  Speaking as someone who has done alot of (rpm) packaging.  It would make it cleaner in the future if you wanted a pyepics and epics packages, as almost all linux distros mandate that packages should not ship their own versions of libraries that other packages ship.  So having an easy setup option to not install the shared libs is a good idea.  But of course, all this only matters if you want to get the packages into upstream! 😄

newville:
An rpm for Pyepics should include libca.so under the pyepics folder, as this PR now includes. That's why we're doing this.

Pyepics is a client library.  Making it depend on epics base is a packaging mistake and will cause errors.  We've seen these errors, and the whole point here is to try to avoid them.

PR #96, opened 2017-09-19
Redux: Lib installation

dchabot:
This PR is a re-opening of #92.

By way of the road less traveled - github does not (easily) permit reopening a closed PR once forced pushes have been made to the closed branch. Seems kinda weak...


tacaswell:
:+1: I like this approach much better!

newville:
@dchabot  Thanks -- this looks good to me.  

IIUC, there are no automated tests for the actual observed "problem" we are trying to solve (location of libca.so when using pip install to system python), so we'll have to actually test that.

I'm still unsure of the need for the NOLIBCA option.  Using PYEPICS_LIBCA at run-time is the better approach.

newville:
@dchabot 

>>  there are no automated tests for the actual observed "problem" we are trying to solve 
>> (location of libca.so when using pip install to system python), so we'll have to actually test that.
>  Test added.

Awesome -- I'm playing with this branch too.   And I also see something weird (maybe different?) atexit too. I'll try to make a fix for the error I'm seeing.

> Take a look at the Travis output for matrix cases where NOLIBCA is undefined. The epics.caget() 
> clauses are throwing on process exit (may be related to #89. Looks similar to that report).

Are you referring to the messages

      File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/epics/ca.py", line 216, in initialize_libca
      File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/epics/ca.py", line 179, in find_libca
      File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/epics/ca.py", line 129, in _find_lib
      File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1220, in resource_filename
      File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 440, in get_provider
      File "<frozen importlib._bootstrap>", line 968, in _find_and_load
      File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 887, in _find_spec
    TypeError: 'NoneType' object is not iterable

If so, these are similar to what I saw at 539831defbd4a941c8c0278655ecb5d07103f285, and which I tried to encourage the op for #89 to explore:  when the new code `ca,_find_lib` is run ***during atexit***, some things may surprisingly by `None`.   With the error happening at ca.py line 129, I'd guess we have to (and currently are not) checking for the case on `inp_lib_name` being `None`.   That's a guess.  Weird things can happen at exit.

tacaswell:
`NOLIBCA` is critical, down stream packagers will want to completely opt-out of pyepics bringing it's own complied binaries.

newville:
@dchabot  Windows 64 is /not/ working.  I'll start investigating.

newville:
@tacaswell @dchabot 

>  NOLIBCA is critical, down stream packagers will want to completely opt-out of pyepics
>  bringing it's own complied binaries.

Uh, why is it a problem to bring along files that can easily be opted-out of using? 

I'm beginning to suspect that this is going to be a problem for this project.

newville:
@dchabot  Sorry for the premature merge: I merged this to master with the Windows interface.  But also: I'm OK with merging this as master because it is clearly very far on the right path.

The Win64 problem was that libCom was not being found (I believe *I* suggested it was not needed, so my bad). Putting this back in seems to work for Windows64, full tests in progress, no problems so far.

newville:
@dchabot  Yeah,think this is a little strange too, and I'm not really sure why this is necessary. I'm pretty sure it is /not/ necessary back with v 3.2.6.  But the awesomeness of git bisect showed it worked before a40be6f1b2e3b391a7f82a2d6a877f1cbd8615ad at not after.

I do not know equivalents for RPATH, patchelf, or install_name_tool for Windows dlls.  I'm surprised ca.dll can't find Com.dll in the same folder, but "find shared library" seems finicky on all systems, and sometimes Windows just makes no sense.

Tangentially-related:  I do not understand why master with all the merges is now failing on Travis.   It looks like everything is OK to me, though I can believe I messed something up.

newville:
> Tangentially-related: I do not understand why master with all the merges
> is now failing on Travis. It looks like everything is OK to me, though I
> can believe I messed something up.
>
> If NOLIBCA is defined, then libraries must be discoverable via tests 1 or
> 3 in ca._find_libs(). The relevant, failing tests here rely on
> PYEPICS_LIBCA to find and load 'libca', but that indicates nothing about
> where 'libCom' resides.
>
> That is why I removed those lines in a40be6f
> <https://github.com/pyepics/pyepics/commit/a40be6f1b2e3b391a7f82a2d6a877f1cbd8615ad>
> .
>


Ah, OK, that sort of makes sense.  I see that `ca._find_libca` is using
'PYEPICS_LIB{}'.format(inp_lib_name.upper())  (didn't remember that).
PYEPICS_LIBCOM should probably default to being "next to" PYEPICS_LIBCA.
That's sort of painful...

The bug you found indicates to me that Windows dlls do not have an RPATH
> mechanism or that the mechanism is broken for the pyepics-packaged dlls.
>

I think there is no such thing as RPATH for Windows dlls.

newville:
On Tue, Sep 19, 2017 at 3:35 PM, dchabot <notifications@github.com> wrote:

> @newville <https://github.com/newville> I can (and have) fixed the tests
> to also export PYEPICS_LIBCOM=/path/to/libCom.so in addition to setting
> PYEPICS_LIBCA. I can submit the PR for that too, if you'd like. However...
>
> Obviously that doesn't address the root of the problem(s).
>
> The success you've experienced previously on Windows could be due to the
> way that data_files was installing the DLLs. So, one solution may be to
> use that mechanism for Windows installations, and use the current
> package_data mechanisms for posix-ish systems (unfortunately this also
> reintroduces the potential for library "conflict").
>

 It's true that it did work with 3.2.7 release.   I'm not sure why it's
different now.

For another python project that needs to load shared libraries, I use a
"share" folder, but I also rely on the fact that the libraries will be
there.   That definitely can also encounter  "shared lib A needs to load
shared lib B" and so I rely on patchelf on Linux and install_name_tool on
OSX, but have not seen a problem on Windows -- I was pretty surprised that
it happened here.

I was kind of hoping for a "uniform" install process, but having a NOLIBCA
install option breaks that concept (it also makes supporting pyepics more
difficult).  So, we'll have to settle for "working", which does have some
merit ;).  With that more modest goal, different install location for the
dlls on Windows would be OK.

And also nuke the explicit loading of libCom/Com.dll...
>
I completely agree that this would be nice!

I will be out much of the rest of the week with spotty email, and won't be
at a Windows64 machine until next Tuesday at the earliest.  It seemed like
you had pretty much solved it all, so I was hoping to get this sorted out
today, but maybe it's OK to go a little slower and more deliberately.

PR #97, opened 2017-09-26
need to preload libCom on Windows

newville:
This may resolve the Travis test failures with NOLIBCA. 

First,  pre-loading of libCom happens only on Windows, where it is necessary, but not on Linux, where it is not necessary, and where it was causing problems with the NOLIBCA option. 

Second, _find_lib() really looks at PYEPICS_LIBCA, but not for PYEPICS_LIBCOM even when looking for 'Com'.  If this envvar is set  but we're really looking for 'Com', it will look for libCom next to libca.

This does work for me on Windows 64bit.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -119,8 +119,13 @@ def _find_lib(inp_lib_name):
     find location of ca dynamic library
     """
     # Test 1: if PYEPICS_LIBCA env var is set, use it.
-    dllpath = os.environ.get(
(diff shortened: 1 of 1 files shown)


newville:
@dchabot  Good, glad I as finally able to catch up to your idea!    I think this is sort of working now, though I'm not sure we've solved the sometimes-segfaults-on-exit problem.  Anyone seeing this?

PR #98, opened 2017-09-26
add AutoSaver class for better repeated saves

newville:
This adds a simple AutoSaver class for save/restore operations.  The AutoSaver class is especially designed for repeated saving of the same PV set, as connections are preserved between saves.

Probably needs documentation!


newville:
docs added.

Issue #100, opened 2017-09-27
Stale rpath in bundled OS X shared libraries


dchabot:
It looks like the rpath $ORIGIN hack was either not applied or incorrectly applied to the distributed OS X dynamic libs (see #80, #81, #82).

Installed pyepics from source (736e65d) via `pip install .` from within an Anaconda environment.

Note the paths beginning, "/Users/Newville/..." in the `OSError` exceptions below.

```python
╭─ [chabot@phase23-58-212]:[~/Dropbox/workspace]
╰─>  python
Python 3.6.1 |Anaconda 4.4.0 (x86_64)| (default, May 11 2017, 13:04:09)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import epics
>>> epics.ca.find_libca()
'/Users/chabot/mc/envs/py3/lib/python3.6/site-packages/epics/clibs/darwin64/libca.dylib'
>>> epics.ca.initialize_libca()
Traceback (most recent call last):
  File "/Users/chabot/mc/envs/py3/lib/python3.6/site-packages/epics/ca.py", line 230, in initialize_libca
    libca = load_dll(find_libca())
  File "/Users/chabot/mc/envs/py3/lib/python3.6/ctypes/__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "/Users/chabot/mc/envs/py3/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/Users/chabot/mc/envs/py3/lib/python3.6/site-packages/epics/clibs/darwin64/libca.dylib, 6):
  Library not loaded: /Users/Newville/local/epics/base-3.15.3/lib/darwin-x86/libCom.3.15.3.dylib
  Referenced from: /Users/chabot/mc/envs/py3/lib/python3.6/site-packages/epics/clibs/darwin64/libca.dylib
  Reason: image not found
Traceback (most recent call last):
  File "/Users/chabot/mc/envs/py3/lib/python3.6/site-packages/epics/ca.py", line 230, in initialize_libca
    libca = load_dll(find_libca())
  File "/Users/chabot/mc/envs/py3/lib/python3.6/ctypes/__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
[...]

PR #101, opened 2017-09-27
FIX: osx rpath in libca

dchabot:
Addresses #100.

Magic sauce was:

```shell
  install_name_tool -change \
  '/Users/Newville/local/epics/base-3.15.3/lib/darwin-x86/libCom.3.15.3.dylib' \
  '@loader_path/./libCom.dylib' libca.dylib
```


newville:
@dchabot Great -- beat me too it

PR #102, opened 2017-09-27
MNT: do attempt disconnect if libca is None

dchabot:
Addresses #89.

Perhaps not so much a fix as a blindfold...


newville:
@dchabot looks good.

PR #106, opened 2017-10-26
Improve batch-getting PVs.

mattgibbs:
This PR improves the epics.caget_many method to accept the same arguments as epics.caget, adds documentation for epics.caget_many, and adds a feature to epics.caget where if you pass it a list or tuple of PV names, it will call epics.caget_many.

A (very basic) unit test for epics.caget_many has been added as well.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -179,9 +179,9 @@ quite good.
 
 ..  function:: caget(pvname[, as_string=False[, count=None[, as_numpy=True[, timeout=None[, use_monitor=False]]]]])
 
-  retrieves and returns the value of the named PV.
+  retrieves and returns the value of the named PV(s).
 
-  :param pvname: name of Epics Process Variable
+  :param pvname: name of Epics Process Variable or list of names.
   :param as_string:  whether to return string representation of the PV value.
   :type as_string:  ``True``/``False``
   :param count:  number of elements to return for array data.
@@ -192,6 +192,9 @@ quite good.
   :type timeout:  float or ``None``
   :param use_monitor:  whether to rely on monitor callbacks or explicitly get value now.
   :type use_monitor: ``True``/``False``
+
+If *pvname* is a list of strings, :func:`caget_many` will be called
+instead with the same arguments.
 
 The *count* and *as_numpy* options apply only to array or waveform
 data. The default behavior is to return the full data array and convert to
@@ -271,6 +274,30 @@ desired::
 Of course,character waveforms are not always used for long strings,  but
 can also hold byte array data, such as comes from some detectors and
 devices.
+
+:func:`caget_many`
+~~~~~~~~~~~~~~~~~~
+
+..  function:: caget_many(pvlist[, as_string=False[, count=None[, as_numpy=True[, timeout=None]]]])
+
+  get a list of PVs as quickly as possible.  Returns a list of values for
+  each PV in the list.  Unlike :func:`caget`, this method does not use 
+  automatic monitoring (see :ref:`pv-automonitor-label`), even for large
+  waveform PVs.
+  
+  :param pvlist: A list of process variable names.
+  :type pvlist:  list of str
+  :param as_string:  whether to return string representation of the PV values.
+  :type as_string:  ``True``/``False``
+  :param count:  number of elements to return for array data.
+  :type count:  integer or ``None``
+  :param as_numpy:  whether to return the Numerical Python representation for array data.
+  :type as_numpy:  ``True``/``False``
+  :param timeout:  maximum time to wait (in seconds) for value before returning None.
+  :type timeout:  float or ``None``
+  
+For detailed information about the arguments, see the documentation for
+:func:`caget`.
 
 :func:`caput`
 ~~~~~~~~~~~~~~~~
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -418,8 +418,7 @@ and the log file is inspected::
 
   get a list of PVs as quickly as possible.  Returns a list of values for
   each PV in the list.  Unlike :func:`caget`, this method does not use 
-  automatic monitoring (see :ref:`pv-automonitor-label`), even for large
-  waveform PVs.
+  automatic monitoring (see :ref:`pv-automonitor-label`).
   
   :param pvlist: A list of process variable names.
   :type pvlist:  ``list`` or ``tuple`` of ``str``
@@ -443,7 +442,7 @@ For detailed information about the arguments, see the documentation for
   put values to a list of PVs as quickly as possible.  Returns a list of ints
   for each PV in the list: 1 if the put was successful, -1 if it timed out.
   Unlike :func:`caput`, this method does not use automatic monitoring (see
-  :ref:`pv-automonitor-label`), even for large waveform PVs.
+  :ref:`pv-automonitor-label`).
   
   :param pvlist: A list of process variable names.
   :type pvlist:  ``list`` or ``tuple`` of ``str``
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -449,11 +449,17 @@ For detailed information about the arguments, see the documentation for
   :type pvlist:  ``list`` or ``tuple`` of ``str``
   :param values: values to put to each PV.
   :type values: ``list`` or ``tuple``
-  :param wait:  whether or not to wait for processing to complete (or time-out) for each put before returning.
-  :type wait:  ``True``/``False``
-  :param connection_timeout:  maximum time to wait (in seconds) for a connection to be established to each PV.
+  :param wait:  if ``'each'``, :func:`caput_many` will wait for each 
+    PV to process before starting the next.  If ``'all'``,
+    :func:`caput_many` will issue puts for all PVs immediately, then
+    wait for all of them to complete.  If any other value,
+    :func:`caput_many` will not wait for put processing to complete.
+  :param connection_timeout:  maximum time to wait (in seconds) for 
+    a connection to be established to each PV.
   :type connection_timeout:  float or ``None``
-  :param put_timeout: maximum time to wait (in seconds) for processing to complete for each PV.
+  :param put_timeout: maximum time to wait (in seconds) for processing
+   to complete for each PV (if ``wait`` is ``'each'``), or for processing
+   to complete for all PVs (if ``wait`` is ``'all'``).
   :type put_timeout: float or ``None``
   
 Because connections to channels normally connect very quickly (less than a
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -286,7 +286,7 @@ devices.
   waveform PVs.
   
   :param pvlist: A list of process variable names.
-  :type pvlist:  list of str
+  :type pvlist:  ``list`` of ``str``
   :param as_string:  whether to return string representation of the PV values.
   :type as_string:  ``True``/``False``
   :param count:  number of elements to return for array data.
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -422,7 +422,7 @@ and the log file is inspected::
   waveform PVs.
   
   :param pvlist: A list of process variable names.
-  :type pvlist:  ``list`` of ``str``
+  :type pvlist:  ``list`` or ``tuple`` of ``str``
   :param as_string:  whether to return string representation of the PV values.
   :type as_string:  ``True``/``False``
   :param count:  number of elements to return for array data.
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -275,30 +275,6 @@ Of course,character waveforms are not always used for long strings,  but
 can also hold byte array data, such as comes from some detectors and
 devices.
 
-:func:`caget_many`
-~~~~~~~~~~~~~~~~~~
-
-..  function:: caget_many(pvlist[, as_string=False[, count=None[, as_numpy=True[, timeout=None]]]])
-
-  get a list of PVs as quickly as possible.  Returns a list of values for
-  each PV in the list.  Unlike :func:`caget`, this method does not use 
-  automatic monitoring (see :ref:`pv-automonitor-label`), even for large
-  waveform PVs.
-  
-  :param pvlist: A list of process variable names.
-  :type pvlist:  ``list`` of ``str``
-  :param as_string:  whether to return string representation of the PV values.
-  :type as_string:  ``True``/``False``
-  :param count:  number of elements to return for array data.
-  :type count:  integer or ``None``
-  :param as_numpy:  whether to return the Numerical Python representation for array data.
-  :type as_numpy:  ``True``/``False``
-  :param timeout:  maximum time to wait (in seconds) for value before returning None.
-  :type timeout:  float or ``None``
-  
-For detailed information about the arguments, see the documentation for
-:func:`caget`.
-
 :func:`caput`
 ~~~~~~~~~~~~~~~~
 
@@ -435,6 +411,29 @@ and the log file is inspected::
     XXX:DMM1Ch2_calc.VAL 2010-03-24 11:56:47.536623 -183.5223
     XXX:DMM1Ch2_calc.VAL 2010-03-24 11:56:48.536434 -183.6832
 
+:func:`caget_many`
+~~~~~~~~~~~~~~~~~~
+
+..  function:: caget_many(pvlist[, as_string=False[, count=None[, as_numpy=True[, timeout=None]]]])
+
+  get a list of PVs as quickly as possible.  Returns a list of values for
+  each PV in the list.  Unlike :func:`caget`, this method does not use 
+  automatic monitoring (see :ref:`pv-automonitor-label`), even for large
+  waveform PVs.
+  
+  :param pvlist: A list of process variable names.
+  :type pvlist:  ``list`` of ``str``
+  :param as_string:  whether to return string representation of the PV values.
+  :type as_string:  ``True``/``False``
+  :param count:  number of elements to return for array data.
+  :type count:  integer or ``None``
+  :param as_numpy:  whether to return the Numerical Python representation for array data.
+  :type as_numpy:  ``True``/``False``
+  :param timeout:  maximum time to wait (in seconds) for value before returning None.
+  :type timeout:  float or ``None``
+  
+For detailed information about the arguments, see the documentation for
+:func:`caget`.
 
 Motivation: Why another Python-Epics Interface?
 ================================================
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -435,6 +435,33 @@ and the log file is inspected::
 For detailed information about the arguments, see the documentation for
 :func:`caget`.
 
+:func:`caput_many`
+~~~~~~~~~~~~~~~~~~
+
+..  function:: caput_many(pvlist, values[, wait=False[, connection_timeout=None[, put_timeout=60]]])
+
+  put values to a list of PVs as quickly as possible.  Returns a list of ints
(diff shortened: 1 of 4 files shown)


mattgibbs:
@dchabot Sure, I can do that!  I'll add more commits to this PR.  Stay tuned...

mattgibbs:
OK, done with my caput_many implementation.  It is a little different than caget_many, in that I have separate timeouts for connection and the put itself, since the expected time for those two can be quite different (usually waiting 60s for a connection is a waste of time, but waiting 60s for a put to finish might be reasonable in some cases).

newville:
@mattgibbs  Thanks for taking this on.  I completely agree that better `caget_many()` and `caput_many()` are very good ideas.  I have a couple of comments or topics we should think about:

First:  I'm willing to be out-voted on this, but I sort of don't like having `caget()` act as `caget_many()` if the argument is a list of pv names.   It seems fragile and  If we have `caget_many()`, we don't really need `caget` to have two cases for its argument.   The thing is that someone will soon want list/tuple to be some other iterator.  But a string is a sequence which will make 'XXX:pv1' ambiguous: a single pv name or sequence of one-character pv names.  Yeah, that's kind of absurd, but I think it's better to avoid the potential ambiguity.
 
Second, for `caput_many`, there are 2 possible meanings of "wait": 
  a) put each PV in order, waiting for each to complete before going to the next.  
  b) put all PVs as soon as possible and wait for them to all complete.

I think both are common enough desires that we should try to accommodate both meanings.   Perhaps keyword args of "wait_all" and "wait_each"?:
     caput_many(pvs, values, wait_all=True) 

will put all and wait for all, while
    caput_many(pvs, values, wait_each=True) 

will put them one after the other.  I don't know that I have a strong preference for the interface, but I think it would be helpful to support both meanings.

Other than those, thanks again -- I think these will be very helpful and used frequently.

mattgibbs:
@newville Regarding caget() act as caget_many(): I don't have especially strong views on this either, but heres my motivation: at least at SLAC, we have a huge community of physicists who use the labCA package with MATLAB, which has a lcaGet function that works with single PVs or lists.  I want to make the transition from MATLAB to Python as painless as possible for these guys.

Regarding the issue with somebody passing an iterator in: you bring up a very good point that list and tuple aren't the only things people might pass.  I'd propose then, that we invert the check:  If pvname is a string, run caget(), if not, assume its something we can iterate over and try to run caget_many().

As for caput_many's wait time, I think I know how to implement wait_all, but I'm not sure I like my idea.  I'll add some more commits to this PR and let you decide if it is reasonable.

newville:
@mattgibbs    

For `caget` acting as `caget_many`:   One of the arguments for using python instead of matlab could be that python is better at getting this sort of thing right. ;).   But also, if `caget` does what `caget_many` does, why have `caget_many` at all?  

Currently, `caget_many` is there when you want many PVs (say, more than 20) as quickly as possible and don't need to re-use the PV connections.   OTOH, a regular `caget` with a single PV name creates and retains a PV object -- that's pretty different.   

I could be persuaded, but `caget` taking either a PV name or an iterable container of PV names seems ugly to me --  both matlab and cothreads do this, eh?   Should we interpret these as a design recommendation or as warnings?

Anyway, having `caget` *act* differently based on whether it is handling one or multiple PVs seems like a very bad idea.  Perhaps `caget` should work with either a PV name or an iterable container of PV names but always create or reuse PV objects, and then have a different function (`caget_many_fast`?) to get many PVs as quickly as possible?  Or maybe we should just beef up to the document of "how to read 5000 PVs as quickly as possible"?

For `caput_many` and the meaning of `wait`:  Putting many PVs and waiting for all to be done seems like a very common use-case to me (more common than "wait for each put to complete before starting the next"). We do this sort of thing all the time: you might want to set many parameters for a detector, start many motors moving, etc.  In these cases, you don't really care which `put` will take the most time, but you do want to wait until all have completed.   That's very different from wait for each `put` to complete.  Both "wait for each" and "wait for all" have uses. 

But also:  for `caput_many`, why use a list/tuple of PV names and list/tuple of values?  Why not use a dictionary (perhaps ordered for python 2.7 and 3.5) of pvname: value pairs?  That seems much better to me.

mattgibbs:
caget/caget_many unification
-----------------------------
Let me make one last argument for `caget` handling both, and if it doesn't convince you, I'll change to stand-alone functions for `*_many`.

I'd wager that the majority of people who use the simple `caget` and `caput` functional interface really don't think/care about the way the connections are managed behind the scenes, so the fact that you get different behavior with lists vs. single PVs will probably go unnoticed usually.  It should be clearly documented, but I don't see it causing anybody significant problems - `caget`/`caput` are the magical, I-don't-care-just-give-me-the-data methods.  They save the user time/effort by making decisions about the best way to achieve the goal of getting data.

My interpretation of the python way of doing things is to try (within reason) to handle whatever the user throws at a method, rather than have several different methods for different use cases.

caput_many and dicts
----------------------

As you allude to, using a dictionary in `caput_many` could cause confusion, as it is unordered.  Maybe you have a sequence of PVs to write, and you want to use wait_each to ensure everything happens in order, for example.

Using an `OrderedDict` is probably the 'right' way to do it, but in practice its pretty clunky.  I usually I already have a list of the PVs (often from a directory service that gives me lists of PVs that match a pattern) and a list of desired values, and shoehorning that into the OrderedDict with a zip(), or populating one by looping over the list, is extra work for not much benefit.

newville:
@mattgibbs I don't think this is about "one last argument". We're just trying to get the
design right.  Hopefully we can come to a consensus.  But, we will be cautious about
changing the API, especially on the simplest interface of `caget` and `caput`.  If the
goal of that interface is to be very friendly to novices, then we want to make sure
existing scripts and expectations about how things work do not break.

Trying to make things easier for novices is a great goal.  A consistent and maintainable
interface is also important -- and might be better for novices.  Users coming from a
different language will have a lot to learn, and making that transition easy may not be
the same as making the library calls look the same as their old language.  There are a
lot of variations in CA wrappings, so trying to emulate one to make transition from that
language easier could make it harder for others.

Many people do rely on `caget(pvname)` caching and reusing CA connections.  I am very
reluctant to change that behavior.  So, if `caget` supports either a single PV name or
an iteratable collection of PV names, I think it should make and cache a full PV
connection for each PV (ie, not call `caget_many`) to be consistent with itself.  With
that condition, why not just use

```python
values = [caget(pvname) for pvname in list_of_pvnames]
```

to handle a source of multiple PV names?  There is definitely something to be said for
"use the language".  For one, `caget` wouldn't have to try to handle all the subtleties in what that iterator/generator might be, and handle exactly one PV name at a time.

But there should be a function that gets many PV values as quickly as possible, without
making full PVs.  To be clear, the speed difference can be enormous. Creating and then
waiting for a PV connection takes 30 ms, whereas you can create and wait for 1000 PVs to
connect simultaneously in far less than 30 s (FWIW, this is a feature of the CA library,
nothing to do with Python).  Anyway, a "get many PV values fast" is a useful function to
have. Currently this is called `caget_many()`.  We could consider renaming this
function, but breaking the API is not done lightly.


For `caput_many`: I think both kinds of `wait` are needed.  For whether to use a dict or
2 separate sequences that implicitly share an order, it may not matter much in practice,
but it also gets to the question of what users expect.  I guess a list of PV names and a
list of Values that are assumed to be correlated would be common.  But, it is hard for
me to imagine how one would get those correlated lists other than from a data structure
that associated the PV names with the values, say from lines in a save/restore text
file or the result of a database look up.  So, maybe the most natural structure would be
a a single list of two-element tuples:

```python
caput_many( [(pvname1, value1), (pvname2, value2), ...])
```

or its close relative, a dictionary:

```python
caput_many( {pvname1: value1, pvname2:, value2} )

```

I don't have a strong preference and don't see any of the options as obviously better or worse than others.

Sorry this is so long.  I think that it might be useful to write some documentation for
the proposed changes to `caget` and `caput` that support either single PVs or multiple
PVs to help figure out what would this interface look like.

mattgibbs:
OK, I have an initial implementation of wait_all/wait_each.  I do it slightly differently: The `wait` argument can be `"all"`, or `"each"` (any other value is interpreted as "don't wait").  This implementation might be overly complex - I am a little paranoid about synchronization when EPICS callbacks change some shared variable.

With the existing testing infrastructure, I have a hard time testing the 'each' and 'all' functionality, so the included tests for these features are pretty thin.  For my own purposes, I made a little pcaspy IOC that simulates a put that takes a while to process.  If you want, I'm happy to include it in the tests/setup directory, but it is yet another thing that would have to be running to make the tests work, and would add a dependency on pcaspy to running some of the unit tests.

mattgibbs:
@dchabot I'd love to hear your thoughts on `caput_many` - I hope that I'm going way overboard, and can completely remove the lock-based stuff.

@newville I agree that 
```
values = [caget(pvname) for pvname in list_of_pvnames]
```
Would be the best case, but I don't know how to make it fast for large numbers of PVs.  Something that would be close is a way to signal to pyepics that you want to perform a set of commands in a batch.  Maybe we could use `with`, like this:
```
with epics.Batch() as batch:
    results = [batch.caget(pvname) for pvname in list_of_pvnames]
print results
```
This would clear up the problem of batched commands stomping on the normal `caget` and `caput` behavior, without introducing much more boilerplate (one extra line), and allowing users to utilize python a little better - you could do whatever you want inside the `with` block to add commands to the batch, and they won't be executed until leaving the block's scope.

newville:
@mattgibbs For caput, there are two different goals here:

  1. get values for a list of PVs because it is convenient to do so.
  2. get values for a large list of PV as fast as possible.

These are really very different, and come down to the question of whether you want to make full, monitored PVs.  With the current master / 3.3.0 branch, option 1 can be done as

```python
values = [caget(pvname) for pvname in list_of_pvnames]
```

and option 2 can be done as 

```python
values = caget_many(list_of_pvnames)
```
Option 1 will definitely be slower the first time, as the PVs have to connect, in sequence.
Option  2 will definitely be faster the first time, but not necessarily faster (and probably slower) if done a second time.  That is, if you will want to get the values later, option 1 is better.

Do we need another variation?  Maybe.   But, what is the problem to be solved? If the goal is to help new users migrating from some other implementation, then option 1 is probably the best approach -- it's the most python-like and will give the richest experience.  Option 2 is really a specialized case of fast one-time processing.

For put_many(), why not use PVs and `pv.put(value, use_complete=True)`, then check `put_complete` property either in sequence or with `all([p.put_complete for p in pv_list])`? 

Sorry if I'm seeming dense, but what are the goals again?

newville:
@dchabot 

> It's unfortunate that PV objects are asymmetric in that they offer put completion, but not get
> completion. Otherwise, this pattern could apply to a batch-oriented get as well. 

get() and put() for PVs are definitely asymmetric.  PVs (usually) auto-monitor their values, so that a `get()` will generally return the latest value -- no new network traffic at all.  You can control "get with completion" for pvs with `PV.get(use_monitor=False)` which will do a real, complete `ca.get()`,   

But `get()` is generally much faster that `put()`.  Or maybe better put:  `get()` is rarely significantly longer than `put()`, but `put()` can easily be much, much slower, and there are many situations where you want to know when a `put()` has actually completed. 

> The Channel Access client library and protocol are fundamentally batch oriented and 
> asynchronous. Maybe this an opportunity for another contribution to pyepics...

What does "fundamentally batch oriented" mean?  I would read that as "done in a standalone script to automate some repetitive task".  I think there are many of us using pyepics for long-running apps. 

>>   Many people do rely on caget(pvname) caching and reusing CA connections. I am very
>>  reluctant to change that behavior.
>
> I agree that caget/caget_many should probably use PV objects, but relying on an 
> implementation's side-effects seems like really bad practice.

Well, creating CA channels is expensive, so many applications using CA rely on and expect users to rely on making and keeping connections for much of the life of the process.  Certainly if you're going to read / write a PVs value multiple time in a program, you'll want to maintain those connections.  

For pyepics, `caget()` and `caput()` are built on top of PV objects, which means that each `caget()`/`caput()` will not create a new connection.  I think abandoning that implementation detail would probably be noticed.

newville:
On Fri, Nov 10, 2017 at 4:29 PM, dchabot <notifications@github.com> wrote:

> @newville <https://github.com/newville>
>
> What does "fundamentally batch oriented" mean? I would read that as "done
> in a standalone script to automate some repetitive task".
>
> By "batch oriented", I mean libca queues client-to-server requests
> internally, dispatching them in bulk for efficiency reasons.
>

It does?  That is not my understanding, which is as a generally
asynchronous client/server subscription mechanism with some slightly
magical network discovering of resources baked in.  But, I could be wrong.

> Dispatching is affected via ca_poll(), ca_pend_event() *et al*.
>

My understanding is that ca_poll() is really not significantly different
from "sleep", and ca_pend_event() is "wait until all outstanding events
have processed".  ca_put(), ca_get(), ca_create_subscription(), and many
other functions do ask the server to reply, but I don't think the polling
functions actually sends any information to the server.  But, I could be
wrong.

Anyway, I think of CA has highly asynchronous and multi-threaded, sending
out lots of events to different machines and processes and with essentially
no concept of "these events happen as an atomic transaction: all at the
same time".  In fact, much of the code I write for interacting with devices
would be way simpler if I could know that when I set a pile of PVs and they
all return that the device will be in a well-defined state *right now* and
not 0.025 or 0.25 seconds later.

> PV objects handle this dispatching transparently to the user. IMO, it's
> one of the most user-friendly aspects of pyepics.
>
> You can control "get with completion" for pvs with
> PV.get(use_monitor=False) which will do a real, complete ca.get(),
>
> The asymmetry I refer to is the fact I can do:
>
> import epicsimport time
>
> pv = epics.PV('test:foo')
> pv.put(2.718, wait=False)
> while not pv.put_complete:
>     print('waiting...')
>     time.sleep(1.0)print('put complete!!!')
>
> But, the PV object interface does not offer the symmetrical, asynchronous
> get operation:
>
> def get_complete_callback(*args, **kwargs):
>     print('args = ', args, ' kwargs =', kwargs)
>
> pv.get(callback=get_complete_callback)
> # do other stuff while "get" completes in a callback...
>
>
OK, I see. Yeah, I guess no one ever asked for a
PV.get_complete_callback().   Maybe it would be useful...
Cheers,

--Matt

mattgibbs:
> what are the goals again?

I guess I'm mostly interested in getting/setting values for a large list of PVs as quickly as possible. The problem to be solved that prompted this pull request is this:

A scanner tool one of my colleagues is writing needs to set a list of values for a scan step, then read ~1000 waveform PVs. Scans are manually triggered via a GUI. Getting a list this size really needs the speed of `caget_many`, iterating over `caget` is too slow.  A lot of software at SLAC is written with this 'collect a bunch of data every once in a while' workflow, where monitoring doesn't give you much benefit because the time between data acquisitions is much larger than the time it takes to wait for a batch of explicit gets.  The client-side load of handling traffic for a thousand monitors is potentially a drawback.

`caget_many`/`caput_many` are great for this type of task, and calling them separately instead of `caget` with a list of PVs is not that big of a problem.  I've removed that feature in the latest set of changes to this PR.

Writing additional features to `caget` and `caput` to also handle collections of PVs, (and maintaining monitored `PV` objects) in a faster way than iterating simply over the collection is a good idea too, but I am less interested in writing that right now.

Regarding `caput_many`, I have re-implemented it using PV objects, and it is indeed much nicer.  Thank you for the advice.

newville:
On Tue, Nov 14, 2017 at 1:43 AM, Matt Gibbs <notifications@github.com>
wrote:

> what are the goals again?
>
> I guess I'm mostly interested in getting/setting values for a large list
> of PVs as quickly as possible.
>

The existing `caget_many()` should be able to handle that OK, I believe.

> The problem to be solved that prompted this pull request is this:
>
> A scanner tool one of my colleagues is writing needs to set a list of
> values for a scan step, then read ~1000 waveform PVs. Scans are manually
> triggered via a GUI. Getting a list this size really needs the speed of
> caget_many, iterating over caget is too slow.
>

Uh, well I would urge caution in that assessment.  `caget_many()` will
certainly be much faster than doing many `caget()` or `PV.get()` in the
case of doing it once and for PVs that are not already connected.   If you
are going to do this more than once and/or PVs are already connected,
`caget_many()` may actually be slower.   That is, the price of establishing
connections and monitor callbacks is a one-time, up-front cost.


> A lot of software at SLAC is written with this 'collect a bunch of data
> every once in a while' workflow, where monitoring doesn't give you much
> benefit because the time between data acquisitions is much larger than the
> time it takes to wait for a batch of explicit gets. The client-side load of
> handling traffic for a thousand monitors is potentially a drawback.
>

That's not my experience at all.  I don't typically monitor 1000s of
waveforms (well, of what size?), but I do monitor thousands of PVs for an
archiving system, and I collect lots of data with long-running scripts and
have several GUI Apps connected to many PVs.

I find monitoring many PVs to be very efficient.  It's not just a matter of
"batch of explicit gets" being faster than the time between data
collections: the get()s for monitored PVs should be nearly instantaneous.
If you're worried about the difference between `many caget()` and
`caget_many()`, I would think that monitored PVs is exactly what you want.

Here's a test scipt that might be of interest.  I have a list of 1000 PVs.
These are all scalar values (so we know what the data size is) and only 10%
change frequently (~5% change at 1 to 10 Hz).  About 15% of the PVs come to
us through a gateway, but all the PVs actually exist and connect.  With
this script:
    import epics
    import time
    pvlist = []
    with open('pvlist.txt') as fh:
        for line in fh.readlines():
            if len(line) > 1:
                pvlist.append(line[:-1].strip())

    t0 = time.time()
    pvs = [epics.get_pv(p) for p in pvlist]

    t1 = time.time()
    print("created  %i PVS:  %.2f sec" % (len(pvs), t1-t0))

    vals = [p.get() for p in pvs]
    print("pv.get() %i PVS:  %.2f sec" % (len(pvs), time.time()-t1))

    for i in range(5):
        print("sleep for 1 second")
        time.sleep(1.0)
        t1 = time.time()
        vals = [p.get() for p in pvs]
        print("pv.get()  %i PVS:  %.2f sec" % (len(pvs), time.time()-t1))

I get the following times:
    created  1000 PVS:  0.27 sec
    pv.get() 1000 PVS:  0.16 sec
    sleep for 1 second
    pv.get()  1000 PVS:  0.01 sec
    sleep for 1 second
    pv.get()  1000 PVS:  0.01 sec
    sleep for 1 second
    pv.get()  1000 PVS:  0.01 sec
    sleep for 1 second
    pv.get()  1000 PVS:  0.01 sec
    sleep for 1 second
    pv.get()  1000 PVS:  0.01 sec

That is, connecting 1000 full PVs with all the callbacks and getting
initial value takes less than 0.5 seconds. Getting those values the second,
third, fourth, or fifth time is essentially free. If you have a long
running script or are going to get the values multiple times, this approach
is definitely what you want.


> caget_many/caput_many are great for this type of task, and calling them
> separately instead of caget with a list of PVs is not that big of a
> problem. I've removed that feature in the latest set of changes to this PR.
>
Yes, this looks good to me.

> Writing additional features to caget and caput to also handle collections
> of PVs, (and maintaining monitored PV objects) in a faster way than
> iterating simply over the collection is a good idea too, but I am less
> interested in writing that right now.
>

If we have `caget_many` and `caput_many` and list comprehension, I'm not
sure this is worth it.

> Regarding caput_many, I have re-implemented it using PV objects, and it
> is indeed much nicer. Thank you for the advice.
>

Yep, your `caput_many()` looks good to me....

newville:
@mattgibbs Sorry, I think I forgot to merge this, getting lost in discussion of when to use `caget_many` vs `caget` and whether `PV`s should have get-callbacks and in other tasks at the beamline.

Anyway, the new `caput_many` and improved `caget_many` are very nice, and improved docs and tests too!  Thanks very much.

PR #108, opened 2017-11-30
Fix MANIFEST.in to correct license file name

synapticarbors:


newville:
@synapticarbors Thanks!

Issue #110, opened 2017-12-05
setup.py and install needs to always install all shared libraries


newville:
As brought up in #109, the current setup.py finds only shared libraries for the current system.  That breaks building a portable source kit with `python setup.py sdist`.

I believe setup.py needs to find and include everything in `epics.clibs`, either all the time (easier, more maintainable) or when building an `sdist` package. I see no harm in always providing all the shared libraries.

In fact, I'd sort of prefer "always" to "unless NOLIBCA was set".  I'm still not sure I understand the reasoning behind NOLIBCA.... Can someone remind me why this is a good idea?

dchabot:
> I'm still not sure I understand the reasoning behind NOLIBCA.... Can someone remind me why this is a good idea?

Perhaps a reading of these threads would help, #90, #81?

newville:
On Dec 5, 2017 09:27, "dchabot" <notifications@github.com> wrote:

I'm still not sure I understand the reasoning behind NOLIBCA.... Can
someone remind me why this is a good idea?

Perhaps a reading of these threads would help, #90
<https://github.com/pyepics/pyepics/issues/90>, #81
<https://github.com/pyepics/pyepics/issues/81>?


Hm, not really.  I see "NOLIBCA is critical" but I do not see why.  I
struggle to imagine a valid use case for it at all, let alone one where it
is important.

Installing the shared libs allows a valid backup to be available.  This
does not have to be used -- it's easy to use another lib.  So installing
with NOLIBCA says "I do not want to install a backup plan that I can easily
not use".

More than that, *why* would someone demand that using the supplied shared
libs is undesirable?  Does it not work? I suspect that the demand that
using a different libca is "critical" comes from a misunderstanding of CA.

Supporting NOLIBCA is not zero work, as this issue demonstrates. So, I
think it is reasonable to br able to justify having this option. At the
moment, I cannot do that.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/pyepics/pyepics/issues/110#issuecomment-349338017>, or mute
the thread

newville:
>>  Hm, not really. I see "NOLIBCA is critical" but I do not see why. I struggle to imagine a valid 
>> use case for it at all, let alone one where it is important.
>
> Are you maybe conflating "not useful to me" with "not useful to others"?

[...]

PR #111, opened 2017-12-05
BUG: install all the libs

dchabot:
Addresses #110 by installing all libraries in `epics.clibs` for all supported architectures, unless `NOLIBCA` is defined. In that case, no libs are installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,8 +44,8 @@ problem before tyring to use the epics package.
 
 nolibca = os.environ.get('NOLIBCA', None)
 if nolibca is None:
-    pkg_data = {'epics.clibs': [epics.utils.clib_search_path("ca"),
-                                epics.utils.clib_search_path("Com")],}
+    pkg_data = {'epics.clibs': ['darwin64/*', 'linux64/*', 'linux32/*',
+                                'win32/*', 'win64/*']}
 else:
     pkg_data = dict()
 


newville:
@dchabot maybe this is a good opportunity to get rid of `pkg_data` completely, and append the platform-specific directories under `clibs` to `packages`, as

     setup(...,
           packages = ['epics','epics.wx','epics.devices', 'epics.compat',
                       'epics.autosave', 'epics.clibs',
                       'epics.clibs.win32', 'epics.clibs.win64',
                       'epics.clibs.linux32', 'epics.clibs.linux64',
                       'epics.clibs.darwin64'] )

If needed (?), I suppose that package list could be built dynamically.

That is, I believe this does not need to be handled as package data.

tacaswell:
The binary libraries should be handled as package data.  

It is absolutely essential that there is a way to opt-out of installing the binary blobs.

newville:
@tacaswell 

> The binary libraries should be handled as package data.

They could be handled as `packages` or as `package_data`.  Using `packages` would mean there would be no package data at all. What's the argument for keeping `package_data` for stuff inside the source tree?

> It is absolutely essential that there is a way to opt-out of installing the binary blobs.

Really?  Absolutely essential?  I am skeptical of that.

I'd like to hear an explanation for why it is ever needed, let alone "important", and never mind "absolutely essential".   I suspect there is a misunderstanding of what installing these shared libraries would mean and how they would be used.

newville:
On Wed, Dec 6, 2017 at 5:36 PM, dchabot <notifications@github.com> wrote:

> What's the argument for keeping package_data for stuff inside the source
> tree?
>
> I don't understand why you want to move it elsewhere. What's the rationale?
>
> There are no modules in epics.clibs - there is nothing to import. It's
> simply not a real package, and shouldn't be construed as one. IMO,
> package_data is the right way to include, distribute, and install
> non-python artifacts.
>
> epics.clibs is in the source tree to be able to encapsulate the libs and
> capitalize on setuptools. This comes down to the eggs vs pip install
> dichotomy. If you can get pkg_resources.resource_filename lookup to work
> outside the tree, go for it. Not sure that gains any advantages though.
>

I do not want to move the final location.  The shared libraries would still
be in site-packages/epics/clibs/darwin64 etc, or in epics/clibs/darwin64
inside the pyepics egg.    That's where they currently go as package data,
and where they would go as part of `packages`.  That is, by having these
file under epics/clibs they can be installed exactly as code, they no
longer need to be installed as data.

Using `packages` instead of `package_data` would just simplify `setup.py`,
but will not change the install location -- in fact it more explicitly
demands that these files must be under siite-packages/epics/clibs where
they will be seen as code.

I don't really have a strong opinion one way or the other.  It's only that
"hey, we could just append those folders to packages and avoid package_data
altogether".  In fact, I'm sort of shocked to see anyone have a strong
opinion on this at all.

I sort of suspect that there may be a misunderstanding of where the shared
libraries get installed with 3.3.0.

I'd like to hear an explanation for why it is ever needed, let alone
> "important", and never mind "absolutely essential". I suspect there is a
> misunderstanding of what installing these shared libraries would mean and
> how they would be used.
>
> Sigh... bait taken...
>

Well, I certainly appreciate you taking the effort to explain this. But, I
am not baiting you, I am asking an honest question and trying to understand
your point of view.  I've asked more than once.   You might do well to tone
down the attitude a bit.  This entire project is a volunteer effort.  I do
a lot of other stuff.  I am sure you do to.  I am trying to keep this
project sane, and to work with you as well as I can.  But if we aren't
communicating well, we are not working together. If you think I am baiting
you here, then there is a problem in our ability to work together.  I'll
assume that we can get over this, but I think you're going to have to work
on this too.  Polite engagement in discussions would be a start.

Here is why NOLIBCA is needed and important (to me):
>
> It's about choice and sound usage patterns.
>
> I want the *option* to use whatever version of libca i see fit for my
> purposes without being forced to use an environment variable to locate my
> library of choice, and without having to opt into whatever version of libca
> is made available with the pyepics bindings.
>

But, you do have the option to use any version of libca you want.  The way
you do that is by setting environmental variables.   Libca is found either
by putting it in a Python-specific place that can be determined from Python
paths, or by setting an environmental variable. That environmental variable
could be PYEPICS_LIBCA, or it could be PATH, or it could be LD_LIBRARY_PATH
/ DYLD_LIBRARY_PATH.  On linux, you can also use ldconfig, but that's sort
of like setting an environment variable system-wide.   There is not and
never has been another way to do it. In fact, I don't see another way to
possibly do it.

I notice that you are *not* saying that the current search process should
be changed.  The current behavior is not completely without reason, but
we've certainly changed it before, and could change it again.  You are most
definitely saying that you want to not even install a known-to-work version
of the required library that you can easily opt out of using. That will
always seem weird to me.

One of the reasons this is important to me is, as explained already, I
> package pyepics for installation via Debian's apt tools. Others have
> created similar packages for epics-base, extensions, etc. Packaging sanity
> dictates that a single version of a library dependency also be available
> for installation *via the same mechanism* (i.e. apt). To satisfy that,
> there must exist *the choice* to opt out of installing the libca version
> (recently) distributed with pyepics, as an existing package already
> provides libca *et al*.
>

Uh, well, maybe.

If I understand, there seem to be two different concerns: a) you want to
explicitly specify which libca to use, and b) you want only one file on
your system named "libca.so".   Those two desires appear to me to have the
potential for conflict.

I don't see why a python package cannot include a shared library that is
hidden away somewhere like site-packages/epics/clibs/linux64/libca.so where
no other package could possibly find it.  There is no potential for
conflict.  There is also no real reason for pyepics to use exactly the same
version of libca.so as any other epics tools -- a feature of CA.  Many
language libraries come with their own share libraries that they install
and use.  Why shouldn't pyepics do this?

I think maybe someone said "one should avoid having 2 files named
libXXX.so" and then someone else interpreted that as "Thou shall not have 2
files with the same name on the same filesystem". Of course lots of
software tools violate that rule.  Epics base is designed to cross compile
and so can certainly generate multiple files named libca.a and libca.so for
different targets and architectures.  I can't imagine how the debian rules
would handle that.  Many linux systems have both /usr/lib/ and /usr/lib64
with many overlapping filenames like lib*.so.

So, I can believe the debian packaging system wants to avoid name clashes.
But I'm not so sure that this is essential for sanity -- the libraries can
be put places where they will be found only by certain packages and will
not be found by other tools.  Like, if a file is installed as
/usr/lib64/python-2.7/site-packages/epics/clibs/libca.so, it would probably
be pretty clear what the intended use is.  It would even be weird to think
that it would be used by Python 3.6 (if/when debian would have packages for
both.).

Now, we could rename the libraries under epics/clibs to ca4py.so or
something.  That might better satisfy the "don't have two files named
libca.so", but it would still require using something like PYEPICS_LIBCA to
specify if you want to use a different library.   Would that be better?  I
don't have a strong preference.

Asking every user to resort to PYEPICS_LIBCA usage to find the right
> library is not a viable alternative, as there are innate, transparent means
> to achieve the desired effect (binutils ld). Using an environment
> variable to locate a library should only be necessary in atypical
> circumstances (eg: testing or development). If this was a good pattern to
> follow, I would have to tell numpy where to find libblas or sqlite3 where
> to find libsqlite. It is not, and I don't.
>

We are most definitely not asking every user to use PYEPICS_LIBCA.  Only
those that do not want to use the provided version of libca.so.  I think
that may be fewer than 5 people, actually.  For sure, using the default
version of libca should be the normal approach.

So, no, you do not have to tell pyepics where to find libca.  You *can*
tell pyepics where to find an alternative version of libca.  That's sort of
unusual, actually. It seems like a useful option, but it is most definitely
an option, and not without costs.

As for libsqlite, in fact I have multiple files named 'libsqlite3.so' on
several linux boxes. Each Anaconda install or virtual environment provides
their own,   For reference, I have 40 files name libsqlite3.dylib on my
Macbook.  Admittedly, I tend to keep a lot of Python environments or
benchmarks.  It turns out that three files name "libsqlite3.dylib" under
/Applications/Xcode.app, all installed at the same time with the same
update from the system package manager, and each of these have unique
checksums.  So, there seems to be some disagreement among package managers
about what "packaging sanity" means.  That's not to say that debian is
wrong, just that it is not the only way to do it.

Anyway, I see a lot of "packaging system XXX has norms that suggest
behavior ABC", which is somewhat different from "absolutely essential".
That's not to say that we can't figure out a way to make a debian package
work better.  But asserting that pyepics must work some way because that's
the way debian wants it to work is definitely not the only concern.

Finally: *Is* there a debian package for pyepics?  I honestly do not know.
I don't see one.  If so, does that even need to install with setup.py at
all?  I don't know anything about making debian packages.

I'm not saying (and have actually never said) that NOLIBCA must be dropped,
only that I don't understand why it is necessary.  I now understand better
what your concerns are, but still don't really see it as necessary, let
alone essential.  It appears to be an attempt to workaround some perceived
problem that is not really that big of a deal.

It also seems that you got yourself into the situation where you think you
need to set PYEPICS_LIBCA when you actually don't *need* to do this, but
*can* do this. That might also be an attempt to solve a perceived problem
that is not actually a very big deal.  It all seems to end up with assuming
(or perhaps demanding)  that you need to use many non-standard options to
use pyepics at all.   There are probably other ways to do it.

tacaswell:
I am with @dchabot 100% here.  At a facility scale where we are centrally managing software deployment (either via system level packages or conda) we want to be able to control what version of libca (and friends) is installed, how they are complied, and what is used at run time.  Ideally this should be possible without special-casing anything in theruntime enviroment for pyepics.  This isn't a hypothetical, at least 3 facilities (FRIB, NSLS-II, and LNLS) care about this. 

I understand how for one-off installations having it 'just work' is convenient, but to make that work the library search is doing some non-standard things which can make integrating pyepics into a broader packaging scheme difficult.  I think the status-quo (modulo this PR) basically satisfies everyone (it 'just works' naively and there is a way to opt-out for people packaging it).

I think the  'correct' solution here would be to make system-dependent wheels so we do not have to include all of the versions on all of the binaries on every system.

newville:
@dchabot @tacaswell  I think you want a special case install that is independent of pyepics.

newville:
@dchabot @tacaswell I also think you are misunderstanding CA.

newville:
@dchabot It occurs to me that the fundamental disagreement is whether a pyepics package (for any system) should depend on an epics-base package.  I think it should not.  It should definitely be possible to install pyepics on a "client only" machine, without epics base.  The good news is that CA is highly tolerant to having multiple CA libs on a system.   Hope that helps clarify my view. 

Of course, if you want to force the same libca for client and server you can do that.  But this is most certainly not necessary.

ralphlange:
Completely agree with @tacaswell and @dchabot.
Add BESSY II to your facility list for the Debian packages. I am packaging pyepics for the CODAC Core System distribution (RHEL based) with ~200 installs at 60+ facilities.

This is not about server and client using the same libca and friends, it is about all tools on any specific machine using the same, managed, known version.

tacaswell:
> @dchabot It occurs to me that the fundamental disagreement is whether a pyepics package (for any system) should depend on an epics-base package. I think it should not.

This is the crux of the confusion.  `pyepics` does depend on having `libca` and `libcom` available, they both come from `epics-base` therefor pyepics depends on epics-base.

@dchabot and I would like this to be clear in the packaging, @newville would like to bundle (parts) of epics-base into pyepics.  Both positions have merits, but this is why `NOLIBCA` is critical to give us the escape hatch we need to easily do the packaging properly.

@newville What part of CA am I misunderstanding?  


> @dchabot @tacaswell I think you want a special case install that is independent of pyepics.

I do not understand this comment at all.


> This is not about server and client using the same libca and friends, it is about all tools on any specific machine using the same, managed, known version.

100% agree with this.

newville:
On Thu, Dec 7, 2017 at 10:28 PM, Thomas A Caswell <notifications@github.com>
wrote:

> @dchabot <https://github.com/dchabot> It occurs to me that the
> fundamental disagreement is whether a pyepics package (for any system)
> should depend on an epics-base package. I think it should not.
>
> This is the crux of the confusion. pyepics does depend on having libca
> and libcom available, they both come from epics-base therefor pyepics
> depends on epics-base.
>

Actually, pyepics does not depend on epics base.

Pyepics is a self-contained install on all supported system, and it will
work without epics base being installed at all. Pyepics comes with its own
versions of all the libraries needed for it to work.  It installs them in a
place only pyepics will find them.  These are the libraries that are tested
and known to work. Using them should absolutely be the default behavior,
and not using them should be considered experimental and not recommended
for production use.

Any packaging system should definitely use the version as normally
installed.

Epics Base is a complex, multi-featured package.  It is designed to build
and cross-compiled libraries for a variety of epics host environments, and
to build epics host IOCs as well as client code.  As it turns out (and this
is by design, but can be considered an implementation detail) the libraries
pyepics uses are exactly the libca and libCom libraries built by Base.
These may be built with specific settings or options to make pyepics work
better.   No other part of Epics Base is ever used by pyepics. Furthermore,
using versions of libca/libCom other than those provided with pyepics may
cause problems on some systems.  As it happens, most such problems are on
linux.

But, anyway, pyepics really does not require Epics Base to be installed.  A
package for pyepics should not depend on a package for epics base.  There
really is no such dependency.

Many extension libraries for interpreted languages (and possibly other
wrappings of CA) would simply ship with shared libraries and not ever
imagine that one would be able to replace them with one built outside the
context of that wrapper.  By using exactly libca/libCom built by epics
base, pyepics allows this to be done fairly easily.  It is possible.  But
it is not ever necessary.

Epics CA has been especially stable in its API, feature set, and
performance.   Pyepics can work with CA libraries going back to at least
3.14.10, though some features (subarrays) do not work with older versions
of libca.  Version 3.14.12 or later is definitely recommended, and the
shipped versions are from the 3.15 series. Of course, any features that
might be introduced in later versions would not be useful in pyepics until
support for them is added.  If the API changes, it would break pyepics.  In
fairness, this seems to never happen. But the counter-argument is also
valid:  there is no feature-driven reason to use a later version of
libca/libCom.

There is absolutely no problem for various epics hosts, clients, and
servers to be using different versions of the CA library, even going back
to 3.13.  Granted, some features may not be available on all machines, but
the protocol and library are deliberately stable and tested across CA
versions so that it can work well in highly diverse and distributed
environments.  We see this in practice every day.

Pyepics is tested with the libraries it ships. Not using these libraries is
throwing away the test results. As above, it seems unlikely that a new
version of libca/libCom will break features in pyepics, but if there are
any changes at all, there is no evidence that they will be for the better.

@dchabot <https://github.com/dchabot> and I would like this to be clear in
> the packaging, @newville <https://github.com/newville> would like to
> bundle (parts) of epics-base into pyepics. Both positions have merits, but
> this is why NOLIBCA is critical to give us the escape hatch we need to
> easily do the packaging properly.
>

No, the NOLIBCA option is definitely not critical. In fact, it says "toss
out the known-to-work installation and replace it with something else".  Is
the debian package tested?  I don't know (I don't actually know if or where
it exists), but I would imagine you do some testing.

I've never created a debian package, but in all the packaging systems I've
ever tried to use, there are always steps of  "pre-install script/patch",
"build/install" (sometimes separated) and a "post-install script/patch".

Any packaging system could definitely build and install pyepics with a
default build and then patch the result. For example, if you wanted to
replace epics/clibs/linux-64/libca.so and epics/clibs/linux-64/libCom.so
with versions from the Epics-Base package, you could do that.

What you do not need is for the upstream package (in this case, pyepics) to
provide a way to break its installation.  And, just to reiterate using
NOLIBCA will create a broken installation of pyepics unless the person
installing with NOLIBCA also installs some otherwise unnecessary package.
NOLIBCA creates the need for a dependency that does not exist.

@newville <https://github.com/newville> What part of CA am I
> misunderstanding?
>

You have said more than once that NOLIBCA is critical.  It is not.  Epics
CA is deliberately tolerant to variations in versions of CA library used.
There is no reason to insist or even expect that all CA tools on a
particular machine or network use exactly the same version of CA
libraries.  Pyepics ships with a tested version of libca.so / libCom.so.
This is well isolated from the rest of the OS and will not be accidentally
found by other tools.

If you *want* to, you can replace the version of libca.so / libCom.so that
pyepics uses.

It is certainly not necessary for an upstream package (here, pyepics) to
support an installation method that will break a normal installation of
that package (as NOLIBCA will) to meet the desires of particular package
manager (here, debian).

@dchabot <https://github.com/dchabot> @tacaswell
> <https://github.com/tacaswell> I think you want a special case install
> that is independent of pyepics.
>
> I do not understand this comment at all.
>

You want to alter your debian build scripts to not use NOLIBCA when
installing pyepics. Find another method.   You could just use the installed
package as is.  This is what I would recommend -- it's the best tested
approach.  You could replace or remove epics/clibs/linux-64/libca.so /
libCom.so in a post-installation step.  You could hardwire-in a
PYEPICS_LIBCA environmental variable with an additional script in
/etc/profile.d.  There are probably other choices.

But you should not be installing with NOLIBCA.  Pyepics should not support
this installation method.  It is never necessary, and it breaks normal
installs.  It is an option that is exactly as useful as 'install this
package, then go through all the files and replace the letter "e" with
"q"'.  It is designed to break the install and must be undone to make the
package useful.

This is not about server and client using the same libca and friends, it is
> about all tools on any specific machine using the same, managed, known
> version.
>
> 100% agree with this.
>

Yes, this is exactly what I am concerned about too.  Pyepics is in a
managed state with a known version, and tested.  Installing with NOLIBCA
undoes that.

newville:
@dchabot OK, this is a simple fix.  I'll merge it.   I never got a good answer for why this still needs to be `package_data`, but it's not a big deal.   I also still do not understand why we need `NOLIBCA` when it is only useful for a downstream packaging system.

PR #115, opened 2018-01-26
Nopoll on ca create conncb

newville:
This PR changes `ca.create_channel` to remove the `poll()` if a connection callback is defined (as is usual).  On Windows64, this call was definitely slow -- taking ~15 ms -- which is not seen on Linux or OSX. 

I'm not sure why that poll is necessary.  That is, it still doesn't ensure that the connection callback has been run before `create_channel` returns.  

FWIW, there is also a change to add more things to the initial list of motor attributes, which is how I started looking into this problem in the first place.   

Travis seems to say this is passing, which is pleasantly surprising -- I half expected a failure when testing connection callbacks.

I'm not sure this needs to be merged immediately.  But, any objections or concerns?


newville:
this seems to be working fine for me after a week of real use and is passing tests, so I'll merge this and wait for reports of it not working ;).

PR #119, opened 2018-02-27
Py3 / wxPython4 fixes

newville:
this fixes a number of small Python 3 / wxPython 4 problems for code in the wx/ folder.  I'm not sure it fixes *all* problems, but at least 1 wxPython4 + pyepics app now works for me.


newville:
merging this -- it seems uncontroversial as it only affects wxPython + Python3

Issue #120, opened 2018-03-17
ca.create_subscription fails on unconnected channel


till-s:
Thanks for pyepics.

I have a mostly asynchronous application where I want to 

ca.create_channel()
ca.create_subscription()

during initialization and not bother with the connection state. I would expect the subscription callback to be executed eventually, when the channel connects.

This works just fine with the native CA library -- but in pyepics ca.create_subscription fails if the channel is not connected. 

I also noted that if I manage to create a subscription then it remains in place across channel dis- and reconnect events - which is great. I just wish the initial connection would be handled more gracefully, i.e., automatically.

till-s:
Obviously my opinion is somewhat biased...
I would indeed argue that an 'ftype' argument would be desirable
 - allows you to subscribe to an unconnected channel
 - allows you to delegate conversion to the desired 'ftype' to epics (which is what it is designed for)
I find it actually cumbersome to always be handed the native type and then have to deal with an appropriate conversion in the callback. E.g., if I have a dumb application which displays stuff in a widget which I need to to update with a string then

        ca.add_subscription(..., ftype=dbr.STRING)

would always hand me a string in the callback, regardless whether the native type is an int, enum or string etc.

Currently, if the PV is natively an 'enum' then I get an integer in the callback. I therefore have to set `use_ctrl=True` *and* check if there are `enum_strs` in the `kwargs` and if there are then I have to map the integer to an enum string. I'd rather have all of that happening behind the scenes and just say `ftype=dbr.STRING`.

This patch (based on your idea) seems to do the job...
[ca.py.diff.gz](https://github.com/pyepics/pyepics/files/1826180/ca.py.diff.gz)

newville:
@till-s  Thanks...  I have incorporated many of these changes.  I'm puzzled by some errors showing up in testing which make it looks like connections are not always completing (or maybe Travis is having trouble...).  I'll check into this more.

But also, you wrote:

>    [adding `ftype` to `ca.create_subscription`] allows you to delegate conversion to the desired 'ftype' to epics (which is what it is designed for)
>    I find it actually cumbersome to always be handed the native type and then have to deal with an appropriate conversion in the callback. E.g., if I have a dumb application which displays stuff in a widget which I need to to update with a string then

>     ca.add_subscription(..., ftype=dbr.STRING)

> would always hand me a string in the callback, regardless whether the native type is an int, enum or string etc.

[...]

PR #121, opened 2018-03-17
allow ca.create_subscription() on unconnected channels

newville:
This tries to address #120, allowing `ca.create_subscription()` on unconnected PVs.

This probably needs documentation and maybe a test or two.  I'm posting this now mostly to demonstrate that this is possible, and might be useful.....


newville:
This is now passing all tests and seems like a nice feature to have.  

Does anyone have any concerns or objections to merging this?

newville:
with no objections, I'm going to merge this.

Issue #116, opened 2018-01-29
Trouble with bundled libca.so.


mattgibbs:
I'm running into some issues with pyepics and libca.  The bundled 32-bit and 64-bit libca fail for me for two different reasons, and if I try to remove those, and use the libca thats in my LD_LIBRARY_PATH, pyepics fails to find that.  (I'm happy to split these up into separate issues if you would prefer, let me know.)

32-bit libca.so
--------------

When using the 32-bit libca.so that comes bundled with pyepics, here's the error I see:

```Python 2.7.4 (default, Jul 22 2013, 22:13:40) 
[GCC 4.1.2 20080704 (Red Hat 4.1.2-54)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import epics
>>> epics.PV('VPIO:IN20:111:VRAW')
Traceback (most recent call last):
  File "/afs/slac/g/acctest/package/python/current/lib/python2.7/site-packages/epics/ca.py", line 227, in initialize_libca
    libca = load_dll(find_libca())
  File "/afs/slac/g/acctest/package/python/current/lib/python2.7/ctypes/__init__.py", line 443, in LoadLibrary
    return self._dlltype(name)
  File "/afs/slac/g/acctest/package/python/current/lib/python2.7/ctypes/__init__.py", line 365, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /afs/slac/g/acctest/package/python/current/lib/python2.7/site-packages/epics/clibs/linux32/libca.so: undefined symbol: _Z26epicsSingletonPrivateMutexv
```

I don't know how to troubleshoot this one further.

64-bit libca.so
---------------

The problem I have with the bundled 64-bit libca is really my own fault: it is built against a newer version of glibc than my (ancient) OS comes with:
```
>>> import epics
>>> epics.caget("B")
Traceback (most recent call last):
  File "/afs/slac/g/lcls/package/python/python2.7.9/linux-x86_64/lib/python2.7/site-packages/epics/ca.py", line 227, in initialize_libca
    libca = load_dll(find_libca())
  File "/afs/slac/g/lcls/package/python/python2.7.9/linux-x86_64/lib/python2.7/ctypes/__init__.py", line 443, in LoadLibrary
    return self._dlltype(name)
  File "/afs/slac/g/lcls/package/python/python2.7.9/linux-x86_64/lib/python2.7/ctypes/__init__.py", line 365, in __init__
[...]

PR #123, opened 2018-06-11
Add another method to find shared libraries using 'ld'

mattgibbs:
This is my attempt at fixing #116 by using the linker to find shared libraries.  This is pretty much back-porting something that ctypes.find_library() does starting in Python 3.6.

To test it:
1. Clear 'PYEPICS_LIBCA' environment variable
2. Remove the bundled libraries in epics/clibs/
3. Ensure libca and libCom are in LD_LIBRARY_PATH (or DYLD_LIBRARY_PATH on Darwin)
4. Run epics.ca.initialize_libca() and see if it finds the library in your LD_LIBRARY_PATH.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -113,6 +113,40 @@ class CASeverityException(Exception):
     def __str__(self):
         return " %s returned '%s'" % (self.fcn, self.msg)
 
+def _findLib_ld(name):
+    """
+    Find dynamic library by asking the linker, then parsing the output, which
+    will list the location where it found the library.
+    
+    This code is based on a patch to the ctypes.find_library() method that
+    was applied in Python 3.6, to fix a problem where it wasn't searching
+    LD_LIBRARY_PATH.  See https://bugs.python.org/issue9998
+    """
+    import re
+    import subprocess
+    expr = r'[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
+    cmd = ['ld', '-t']
+    libpath = os.environ.get('LD_LIBRARY_PATH')
+    if sys.platform == 'darwin':
+        dylibpath = os.environ.get('DYLD_LIBRARY_PATH')
+        if dylibpath is not None:
+            libpath += ":" + dylibpath
+    if libpath:
+        for d in libpath.split(':'):
+            cmd.extend(['-L', d])
+    cmd.extend(['-o', os.devnull, '-l%s' % name])
+    result = None
+    try:
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE,
+                             universal_newlines=True)
+        out, _ = p.communicate()
+        res = re.search(expr, out)
+        if res:
+            result = res.group(0)
+    except Exception as e:
+        pass  # result will be None
+    return result
 
 def _find_lib(inp_lib_name):
     """
@@ -174,6 +208,11 @@ def _find_lib(inp_lib_name):
     # with PATH set above, the ctypes utility, find_library *should*
     # find the dll....
     dllpath = ctypes.util.find_library(inp_lib_name)
+    if dllpath is not None:
+        return dllpath
+
+    # If we still didn't find it (which is likely in Python<3.6), use a method back-ported from Python 3.6.
+    dllpath = _findLib_ld(inp_lib_name)
     if dllpath is not None:
         return dllpath
 
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -212,9 +212,10 @@ def _find_lib(inp_lib_name):
         return dllpath
 
     # If we still didn't find it (which is likely in Python<3.6), use a method back-ported from Python 3.6.
-    dllpath = _findLib_ld(inp_lib_name)
-    if dllpath is not None:
-        return dllpath
+    if os.name == 'posix':
+        dllpath = _findLib_ld(inp_lib_name)
+        if dllpath is not None:
+            return dllpath
 
     raise ChannelAccessException('cannot find Epics CA DLL')
 


newville:
@mattgibbs it's been awhile since I thought about this, so you may have to remind me (and maybe others) of the essentials of the problem and the strategy here.  If I recall, the issue was with very old Python version and a very old glibc/gcc. 

Opening a subprocess to `ld` seems kind of error-prone to me.  I guess you are saying that explicitly walking through PATH + LD_LIBRARY_PATH + DYLIB_LIBRARY_PATH in Python does not work.  Weird.

I'm not sure why the right answer is not "use PYEPICS_LIBCA".   I think that any reliable solution to a mismatch of shared libraries on POSIX systems *must*  involve setting an environmental variable.  Setting PATH or LD_LIBRARY_PATH or DYLIB_LIBRARY_PATH could work but these variables are definitely used by other applications, and so might cause conflicts with other apps  (and to be clear, the reason we include libca.so for Linux is that we did really have such a conflict).   Using PYEPICS_LIBCA is explicit, specific, and applies only to how Pyepics finds libca.  Why is that not the preferred solution?

mattgibbs:
Hi @newville, the core of this problem is that PyEPICS doesn't search LD_LIBRARY_PATH on older versions of Python (not very old, anything <3.6, I believe).  I agree that PYEPICS_LIBCA is a fine way to make it work, but I think its reasonable to expect that PyEPICS would behave like most software when doing dynamic linking, and attempt to find the library in LD_LIBRARY_PATH.  The documentation advertises that it does.

When I was trying to figure out why this problem was happening in the first place, I stumbled on to the Python issue/patch that fixes it for 3.6.  Rather than come up with a new solution, it seemed better to just do what they did.  But I will admit: I don't know why they don't just walk every directory in LD_LIBRARY_PATH and look for 'lib{name}.{so | dylib}'.  I guess using 'ld' will also check the OS's library paths (like /lib) too, but for PyEPICS, we can probably always ignore those.

newville:
@mattgibbs 

> Hi @newville, the core of this problem is that PyEPICS doesn't search LD_LIBRARY_PATH on older versions of Python (not very old, anything <3.6, I believe).

Well, PyEpics could explicitly search LD_LIBRARY_PATH. Currently, it does this:
  
   1. check if PYEPICS_LIBCA is set, if so use that. We're done.
   2. check if the appropriate lib was installed in the `epics.clibs` folder.  If so, use that. We're done.
   3. set PATH to include [DY]LD_LIBRARY_PATH, and then use `ctypes.util.find_library()`.  

So, we could change Step 3 to not rely on `ctypes.util.find_library` but to explicitly look for the appropriate library file. 

But, I think that isn't really the problem here.  The problem is that option 2 succeeds with a DLL that is not compatible with your version of glibc.  That version of the DLL *could* be replaced, or you could just delete it and rely on the modified Step 3.

> I agree that PYEPICS_LIBCA is a fine way to make it work, but I think its reasonable to expect 
> that PyEPICS would behave like most software when doing dynamic linking, 

Suggestions for how to do that are most welcome.  I think that Python does not autoload shared libraries if they are defined with ldconfig.  Maybe there is a way to do that.
 
> and attempt to find the library in LD_LIBRARY_PATH. The documentation advertises that it does.

Well, it does find the library, just not the one you want....  

> I guess using 'ld' will also check the OS's library paths (like /lib) too, but for PyEPICS, 
> we can probably always ignore those.

I am not excited about supporting running a subprocess of whatever `ld` happens to be found.  I have strong doubts that this will add reliability.

Again, the problem we saw that prompted us to distribute libca.so on Linux, and the problem you are reporting here are *not* that libca.so was not found, but that an incompatible libca.so was found.   I don't know how to solve that well without setting an environmental variable.   One could set LD_LIBRARY_PATH (and modify the code to work with all Python versions), but I think that setting PYEPICS_LIBCA will always be more reliable.

mattgibbs:
@newville OK, I'll just continue to use PYEPICS_LIBCA in the future.  Feel free to close this PR.  You may want to add a note to the documentation (http://pyepics.github.io/pyepics/installation.html#getting-started-setting-up-the-epics-environment) that LD_LIBRARY_PATH will not be searched unless you are using Python 3.6.

klauer:
Somewhat related: is it worth considering that pyepics could rely on https://pypi.org/project/epicscorelibs/ in place of building and packaging its own?

newville:
@klauer That's an interesting suggestion.  I had not seen this or tried it out those wheels (but from essentially an epics-base repo?), but it does look like an interesting project that seems to include shared libs for all supported platforms, and includes bundling of shared libs for PVAccess as well.   So: could be really good, definitely worth looking in to. 

FWIW, I am strongly in favor of being able to ship what is tested, and so prefer either bundling shared libs or having such a repo with pre-built known-to-work versions.  It's not 100% clear to me how all those wheels were configured and built.   It might be nice to know the version of base used and also to have support for linux-arm (raspberryPi).

Also, recall that the thing that caused us to switch to pre-built shared libs for linux (and the only one that anyone has ever complained about or run into trouble with) was libCom that was linked with a version of libreadline that was incompatible with the one used by Python.   Pyepics does not ever need libCom to include libreadline, so I would want to be cautious about that.

Anyway, that is definitely worth exploring.

klauer:
@newville definitely agreed on shipping what's been tested and proven to work!

I should have pasted the repo link as well (from the tech talk post): https://github.com/mdavidsaver/epicscorelibs

It's probably not worth the effort at this point, but may be worth considering down the line.

newville:
@klauer @mattgibbs I'm going to close this.  

FWIW, the master branch (soon to be 3.3.2) now has shared libs for linux that are built with an old-ish glibc (2.12) and gcc (4.4.7), so it ought to work on more systems (I believe all supported major versions the Redhat/Centos family and the Debian family, not sure about others).  But we can still consider using epicscorelibs as a dependency, or pulling compiled sharedlibs from the Jenkins CI results.

Issue #124, opened 2018-06-25
Severity Update


hhslepicka:
Hi,

I am facing an issue in which I don't get a monitor event with PyEPICS if just the severity of the PV changes.
If the value and severity change everything goes as expected but if the value remains the same and the severity is changed the callback is never executed.

In order to demonstrate that I created this little database:

```bash
slepicka@dhcp-swlan-public-81-101:~/sandbox/iocs/alarms  $ cat test.db
record(ai, "TEST:A")
{
	field(DESC, "VALUE A")
	field(EGU, "Counts")
	field(INP, "TEST:calc.VAL NPP NMS")
	field(HOPR, "5")
	field(LOPR, "0")
	field(HIHI, "5")
	field(HIGH, "3")
	field(LOW, "1")
	field(LOLO, "0")
	field(HHSV, "MAJOR")
	field(HSV, "MINOR")
	field(LSV, "MINOR")
	field(LLSV, "MAJOR")	
	field(PINI, "YES")
	field(VAL, "0")
}

record(calc, "TEST:calc")
{
	field(DESC, "Counter")
	field(SCAN,"1 second")
	field(FLNK, "TEST:A")
	field(CALC, "(A<B)?(A+C):D")
	field(INPA, "TEST:A.VAL  NPP NMS")
	field(INPB, "5")
	field(INPC, "1")
	field(INPD, "0")
}


record(calc, "TEST:calcA")
{
	field(DESC, "Counter")
	field(INPA, "TEST:A CP MS")
	field(CALC, "0")	
}
```

**Launching with softIoc:**
```bash
slepicka@dhcp-swlan-public-81-101:~/sandbox/iocs/alarms  $ softIoc -d test.db
Starting iocInit
############################################################################
## EPICS R7.0.1.2-DEV
## EPICS Base built Apr 25 2018
############################################################################
iocRun: All initialization complete
epics>
```

Via `camonitor` I can see the updates but via PyEpics I can't...

**Output of `camonitor`:**
```
slepicka@dhcp-swlan-public-81-101:~/sandbox/epics/R7.0.X  $ camonitor TEST:calcA
TEST:calcA                     *** Not connected (PV not found)
TEST:calcA                     2018-06-25 11:32:59.135703 0 LINK MINOR
TEST:calcA                     2018-06-25 11:33:00.135888 0
TEST:calcA                     2018-06-25 11:33:01.131490 0 LINK MINOR
[...]

PR #127, opened 2018-07-26
BUG: set default subscription mask to VALUE|ALARM

dchabot:
Address issue #124.

attn @hhslepicka


newville:
Looks good to me.  Anyone have any concerns or objections to this?

hhslepicka:
Awesome, @dchabot and @newville .
Thank you for looking into that.

newville:
OK, I think there is no objection, so I'm merging this.  Thanks @dchabot !

Issue #126, opened 2018-07-26
Char waveform transforms fails if count >= 65536


dchabot:
Attempting to `get` on a waveform PV of type char fails if the NELM is >= 2**16.
The failure manifests as the returned value being the diagnostic string, `<array size=%d, type=%s>`.

The offending line is [here](https://github.com/pyepics/pyepics/blob/1f795c647458435cca0cf11b984b34fc11180c25/epics/pv.py#L413), in `PV._set_charval()`.
I'm having trouble seeing the reason for the necessity of the check that, `self.count < ca.AUTOMONITOR_MAXLENGTH`.

newville:
@dchabot the basic idea was that a CHAR waveform that is not too long may very well be used to store a long string -- this is pretty common usage.  On the other hand, a CHAR waveform that is very long probably is storing an 8-bit image from a camera.   For sure,  that transition between "not too long" and "very long" is somewhat arbitrary, but having it be somewhere between 2**8 and 2**32 does not seem too crazy.   I'm not sure that we want the character representation of a 2**32 image being [chr(x) for x in val.tolist()] there, but I guess that could be argued to be more consistent.....  I don't have a strong preference either way.

dchabot:
> the basic idea was that a CHAR waveform that is not too long may very well be used to store a long string -- this is pretty common usage. On the other hand, a CHAR waveform that is very long probably is storing an 8-bit image from a camera.

Yeah, I suspected this limitation had to do with images.
That's a reasonable thing to do for that scenario.

The trouble arises when I _explicitly_ ask for a string: `epics.PV.get(as_string=True, use_monitor=False)`. The assumption being that the user knows what they're doing. This currently fails, of course, because of the conditional on `self.count < ca.AUTOMONITOR_MAXLENGTH`.

One simple workaround that comes to mind is to add a bypass for this explicitness on the check against `ca.AUTOMONITOR_MAXLENGTH` iff `as_string=True`, by adding an extra argument to `epics.PV._set_charval`:

```python
def _set_charval(self, val, call_ca=True, as_string=False):
    :
    :
    if ntype == dbr.CHAR and (self.count < ca.AUTOMONITOR_MAXLENGTH or as_string is True):
        return convert_waveform_to_string(val)
    :
    :
```

The goals being maintenance of backwards compatibility whilst respecting explicit requests for string representation.

newville:
[...]

PR #128, opened 2018-07-27
ENH: respect formatting iff as_string is True

dchabot:
Addresses issue #126


newville:
@dchabot  looks good to me.  Anyone have any objects to this fix?  I don't expect this will break anything unexpected so if there aren't any concerns, I'll plan on merging by the end of the week.

PR #130, opened 2018-08-03
Deal with timeouts the same way in caget, caput and camonitor

### LINKED ISSUE #129: caget timeout not working as expected
It looks like `caget` timeout is not working when a non-existent PV is involved:

``` python
>>> import timeit
>>> timeit.timeit('import epics; epics.caget("bogus_pv", timeout=0.1)', number=5)
cannot connect to bogus_pv
cannot connect to bogus_pv
cannot connect to bogus_pv
cannot connect to bogus_pv
cannot connect to bogus_pv
10.001068115234375
```

``` python
>>> import epics
>>> epics.__version__
'3.3.1'
```

brunoseivam:
Fix #129

### DIFF
--- a/epics/__init__.py
+++ b/epics/__init__.py
@@ -61,13 +61,16 @@ def caput(pvname, value, wait=False, timeout=60):
     to wait for pv to complete processing, use 'wait=True':
        >>> caput('xx.VAL',3.0,wait=True)
     """
-    thispv = get_pv(pvname, connect=True)
+    start_time = time.time()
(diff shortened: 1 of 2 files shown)


newville:
@brunoseivam Thanks very much.  I think this looks really good.  Looks like @dchabot agrees. 

Does anyone have any objections to merging this?  If not I'll plan on merging soon.

But first (like, now), I'll merge #128.  That might need to be merged with this (as both change  `pv.py`).

newville:
@brunoseivam looks like no one will object, and that this will merge cleanly, so let's just do it.

PR #131, opened 2018-08-07
Fix wait_for_connection timeout

brunoseivam:
Still related to #129, I missed this bit that also had to be changed. Now everything seems to work as expected.


newville:
Ok, that looks fine to me... Iet's just go for it....

PR #135, opened 2018-11-26
ENH: on PV.get(), update all CTRL/STS/TIME metadata

klauer:
Goal
----
* Update `epics.PV` metadata when `PV.get()` is called. 
* Allow access to metadata and value from a specific get request through `PV.get_with_metadata()`, where a kwarg can indicate a request for different DBR types: `ctrl`, `time`, or `None` (default to PV.form)

API additions were needed to make this possible, but it should all be backward-compatible. The original `epics.ca.get` and `epics.ca.get_complete` just now call the 'metadata-rich' versions `epics.ca.get_with_metadata` and `epics.ca.get_complete_with_metadata`.

Metadata update example
-------------
If `pv.form` is set to 'time', then timestamp, status, severity will all be updated before returning. This means that `pv.timestamp` now accurately reflects the timestamp just received from the get request.

Similarly, if `pv.form` is 'ctrl', then all control metadata should be updated on each `.get()` call. 

`PV.get_with_metadata` example
----------
```python
In [1]: import epics

In [2]: mtr1 = epics.PV('sim:mtr1', form='native')

In [3]: mtr1.get_with_metadata(form='ctrl')
Out[3]:
{'upper_disp_limit': 100.0,
 'lower_disp_limit': -100.0,
 'upper_alarm_limit': 0.0,
 'upper_warning_limit': 0.0,
 'lower_warning_limit': 0.0,
 'lower_alarm_limit': 0.0,
 'upper_ctrl_limit': 100.0,
 'lower_ctrl_limit': -100.0,
 'precision': 3,
 'units': 'deg',
 'status': 0,
 'severity': 0,
 'value': 0.0}

In [4]: mtr1.get_with_metadata(form='time')
Out[4]:
{'status': 0,
 'severity': 0,
 'timestamp': 1543429156.811018,
 'posixseconds': 1543429156.0,
 'nanoseconds': 811018603,
 'value': 0.0}

In [5]: mtr1.get_with_metadata(form='time', as_string=True)
Out[5]:
{'status': 0,
 'severity': 0,
 'timestamp': 1543429156.811018,
 'posixseconds': 1543429156.0,
 'nanoseconds': 811018603,
 'value': '0'}
```

Improves
----------
* The case where `auto_monitor=False` - which requires separate calls to `get_timevars` or `get_ctrlvars` to update metadata
* `pv.get(with_ctrlvars=True)` when `pv.form='ctrl'` does not make 2 get requests
* Less code-duplication in `epics.ca`, moves all the metadata unpacking to one function (`epics.ca._unpack_metadata`)

### DIFF
--- a/doc/ca.rst
+++ b/doc/ca.rst
@@ -276,9 +276,13 @@ keyword arguments can be used to specify such options.
    function.
 
 
+.. autofunction:: get_with_metadata(chid, ftype=None, count=None, as_string=False, as_numpy=True, wait=True, timeout=None)
+
 .. autofunction:: get_complete(chid, ftype=None, count=None, as_string=False, as_numpy=True, timeout=None)
 
    See :ref:`advanced-get-timeouts-label` for further discussion.
+
+.. autofunction:: get_complete_with_metadata(chid, ftype=None, count=None, as_string=False, as_numpy=True, timeout=None)
 
 .. autofunction::  put(chid, value, wait=False, timeout=30, callback=None, callback_data=None)
 
--- a/doc/pv.rst
+++ b/doc/pv.rst
@@ -85,7 +85,7 @@ callbacks to be executed when the PV changes.
 
 ..  _pv-get-label:
 
-.. method:: get([, count=None[, as_string=False[, as_numpy=True[, timeout=None[, use_monitor=True]]]]])
+.. method:: get([, count=None[, as_string=False[, as_numpy=True[, timeout=None[, use_monitor=True, [with_ctrlvars=False]]]]]])
 
    get and return the current value of the PV
 
@@ -120,9 +120,45 @@ callbacks to be executed when the PV changes.
    automatically monitored.   Otherwise, the most recently received value
    will be sent immediately.
 
+   The *with_ctrlvars* option requests DBR_CTRL data, including control limits,
+   precision, and so on, in addition to the value normally returned.  This metadata
+   will be available by accessing various attributes such as
+   ``lower_ctrl_limit``.
+
    See :ref:`pv-automonitor-label` for more on monitoring PVs and
    :ref:`advanced-get-timeouts-label` for more details on what happens when
    a :func:`pv.get` times out.
+
+
+.. method:: get_with_metadata([, form=None, [count=None[, as_string=False[, as_numpy=True[, timeout=None[, use_monitor=True, [with_ctrlvars=False]]]]]]])
+
+   Returns a dictionary of the current value and associated metadata
+
+   :param form:  EPICS *data type* to request:  the 'native', or the 'ctrl' (Control) or 'time' variant. Defaults to the PV instance attribute ``form``.
+   :type form:  {'native', 'time', 'ctrl', None}
+   :param count:  maximum number of array elements to return
+   :type count:  integer or ``None``
+   :param as_string:  whether to return the string representation of the  value.
+   :type as_string: ``True``/``False``
+   :param as_numpy:  whether to try to return a numpy array where appropriate.
+   :type as_string: ``True``/``False``
+   :param timeout:  maximum time to wait for data before returning ``None``.
+   :type  timeout:  float or ``None``
+   :param use_monitor:  whether to rely on monitor callbacks or explicitly get value now.
+   :type use_monitor: ``True``/``False``
+
+   See ``PV.get``, above, for further notes on each of these parameters.
+
+   Each request to EPICS can optionally contain additional metadata associated
+   with the value.  While ``PV.get`` updates the PV instance with any metadata,
+   ``get_with_metadata`` will return the requested metadata and value in a
+   dictionary. 
+
+   The exception is when the PV is set to auto-monitor and the `use_monitor`
+   parameter here is set.  This means that both the value and metadata will
+   used the cached values instead of making a new request.  Because of this,
+   the metadata and value returned here will be a full dictionary of all known
+   metadata for the PV instance.
 
 
 .. method:: put(value[, wait=False[, timeout=30.0[, use_complete=False[, callback=None[, callback_data=None]]]]])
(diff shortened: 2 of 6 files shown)


klauer:
Tests pass except a couple unrelated 3.6 failures due to `multiprocessing.Pool` API changes upstream. It may have been broken even before 3.6, but it only starts failing with that version.

Those API changes will make it difficult to support the API going forward, so you may want to consider deprecating/removing support for it altogether (see https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context). A topic for another issue/PR, nonetheless...

klauer:
@newville Great! OK, I've added some docstrings and made at least an `autofunction` in the docs.

I've set it to skip the CAPool tests on 3.x. It's actually a changed API for `multiprocessing.Pool` and its new usage of `multiprocessing.Context` (separate from context managers, though we're using those for the tests). Coming back to it later is probably a wise decision, I think.

newville:
@klauer I think that `PV.get()` already does get the time and control metadata, and stores them in PV attributes and in the (not so well named) `PV._args` dict.   That could be used for a str.format() (which could really simplify `PV._getinfo`!),  perhaps based something like:

```python
pv = epics.PV('XXX:motor1.VAL')
print("""{pvname}: Value={value}, Units='{units}', 
           Precision={precision}, Access='{access}',
           TimeStamp{timestamp}""".format(**pv._args))
```

Is that good enough?  Maybe `_args` should be renamed or aliased to be `metadata`, or `metadata` could be a **property** that fetches `_args`?

klauer:
Right - it does store it there after this PR. Consider the following race condition to see why I think it's still not ideal:

```python
pv = epics.PV('Py:ao', form='time', auto_monitor=True)
value = pv.get()  # <-- _args['status'] is updated
# *** channel access monitor fires in the background, `_args` is completely updated ***
print("""{my_value} {pvname}: args value={value}, Units='{units}', 
           Precision={precision}, Access='{access}',
           TimeStamp{timestamp}""".format(my_value=value, **pv._args))
```
At the final step, the the metadata from `pv._args` is out of sync with `value` that I requested from `.get()` due to the monitor event that happened in-between.

I think it should be possible to request just the value and associated metadata from the specific `get()` request, not including any monitor update information.

newville:
@klauer Ah, yes there is the potential to get out-of-sync.   Maybe a `metadata` property should fetch the STS / CTRL data, even if slightly slower?   Anyway, yes I agree with having a way to get the data **and** force the metadata to be updated.     I think that `PV.get(..., with_ctrlvars=True)` will do a `PV.get_ctrlvars()`, but the API for that might be improved.

klauer:
I think there's some benefit to allowing access to `PV._args`, aliased as `PV.metadata`. Some of that benefit is negated by the magic behind the PV properties using `epics.PV._getarg` that automatically calls `get_ctrlvars` for you when necessary. So the metadata dictionary might not always contain what the user is expecting or hoping for.

I see what you mean about `with_ctrlvars`, but I'm not sure `PV.get` could be modified in a backward-compatible way to get a synchronized set of value and metadata. Do you see a way it could work?

It's easy enough to add the function I proposed above `get_with_metadata` - which wouldn't change the usage of the current API at all. Can you tell me what you don't like about this approach?

newville:
@klauer  I don't have a strong opinion against `get_with_metadata()`.  I started out wondering if maybe it could be an option to `get()`.  That is, `get(with_ctrlvars=True)` already exists, so maybe a similar option to make sure the metadata is up-to-date?   But, adding a `metadata` property or a `get_with_metadata()` method is OK too.

klauer:
@newville I think I see better where you're coming from now. 

My main issue adding to `get()` itself is that I want to change the return type to a dictionary like so: `PV('pv', form='time').get(with_metadata=True) -> {'value': 0, 'status': ..., 'severity': ..., ...}`. While it's technically doable without adding another function, I'd think the return type confusion will cause a lot of problems with end-users. So, that led me down the path of thinking a new function would be better.

Since it seems OK with you, I'll give it a go and let you be the judge after you see some real code.

newville:
@klauer yeah, that is a good point. If the idea is to return a dict, then it should probably be a separate function.

klauer:
OK, I pushed my proposed addition. Here's an example of what it looks like:

```python
In [1]: import epics

In [2]: mtr1 = epics.PV('sim:mtr1', form='native')

In [3]: mtr1.get_with_metadata(form='ctrl')
Out[3]:
{'upper_disp_limit': 100.0,
 'lower_disp_limit': -100.0,
 'upper_alarm_limit': 0.0,
 'upper_warning_limit': 0.0,
 'lower_warning_limit': 0.0,
 'lower_alarm_limit': 0.0,
 'upper_ctrl_limit': 100.0,
 'lower_ctrl_limit': -100.0,
 'precision': 3,
 'units': 'deg',
 'status': 0,
 'severity': 0,
 'value': 0.0}

In [4]: mtr1.get_with_metadata(form='time')
Out[4]:
{'status': 0,
 'severity': 0,
 'timestamp': 1543429156.811018,
 'posixseconds': 1543429156.0,
 'nanoseconds': 811018603,
 'value': 0.0}

In [5]: mtr1.get_with_metadata(form='time', as_string=True)
Out[5]:
{'status': 0,
 'severity': 0,
 'timestamp': 1543429156.811018,
 'posixseconds': 1543429156.0,
 'nanoseconds': 811018603,
 'value': '0'}
```

klauer:
Sorry about that, some dumb oversights that I would have caught if I ran the test suite locally. I'm up for any additional changes/improvements you might want, @newville.

(@dchabot @tacaswell @zllentz @prjemian @trendahl - feel free to chime in, as I know you're using pyepics)

newville:
@klauer no worries from me.   That's what the Travis tests are for! (I should really have a docker test system set up locally, but currently do not). 
I think this is a nice addition and have no request for changes.

newville:
@teddyrendahl  We talked about something like `get(with_metadata=True)`, but decided that `get()` should sort-of-kind-of return a predictable type.  Namedtuples are really nice, but the dict will make it super-easy to use the result with `str.format()`, which was part of the motivation.

Perhaps `get_metadata(as_dict=True)` and  `get_metadata(as_namedtuple=True)` would be reasonable?  I sort of think using a property that was read-only and really fetched the latest metadata values might be useful here too, maybe `metadata_dict` and/or `metadata_namedtuple`.   I do not have a strong opinion on any of this, just tossing out suggestions.

klauer:
@teddyrendahl The only issue with `namedtuple` is that there are different `DBR_CTRL` attributes with different names, so there would be maybe 5+ we'd have to define and use based on the DBR type. If you want an object like that, I'd say we could return a `SimpleNamespace` very easily, though the repr for it is far from ideal.

But why is mutability an issue here? It's a new `dict` for each request.

Changing the return type of a function completely based on a keyword argument is generally very bad practice - see the discussion above about that.

newville:
@teddyrendahl  Though not the same as a namedtuple (and not immutable), for 'dot and tab accessible'  you would be able to convert the dict returned by `PV.get_with_metadata()` to a namespace (Python3 only):

```python
from types import SimpleNamespace
from epics import PV
pv1 = PV('xxxx1')

meta = SimpleNamespace(pv1.get_with_metadata())
print(meta.severity, meta.status)
print("PV : value=%s at timestamp=%f" % (repr(meta.value), meta.timestamp))
```

klauer:
I had a chat with @teddyrendahl offline and realized that perhaps my strong opinions about return types mainly applied to further modifying `get()`. So, I've pushed one additional change that allows for `get_with_metadata` to take `as_namespace`, returning a dynamically-generated namespace in place of the `dict`.

```python
>>> pv = epics.get_pv('sim:mtr1', auto_monitor=False, form='time')
>>> pv.get_with_metadata(as_namespace=True)
namespace(nanoseconds=942818226, posixseconds=1544038869.0, severity=0, status=0, timestamp=1544038869.942818, value=-2.0)
>>> pv.get_with_metadata(as_namespace=True).value
-2.0
```

Unless there are further suggestions, I think this is ready.

newville:
@klauer I agree with you about having a consistent return type from `get()` but allowing some variation with `get_with_metatdata()`: namespaces and dicts both have their uses. 

I think this may be ready to merge, and is going to be a very nice feature.  
Do you think this needs more time or discussion? 

Does anyone else have any concerns?

klauer:
Thanks, @newville! It's ready from my perspective.

newville:
@klauer OK, great.  I'll plan on merging this in a couple of days, just to give anyone who might be interested a little time to raise any concerns.

newville:
@klauer Thanks, merged!

PR #139, opened 2018-12-19
use nelm to determine if a PV is an array, not count: allows 1-element arrays

newville:
This uses NELM>1 to decide if a PV is an array, not `count > 1`.  

This is useful for array PVs that may sometimes contain only one element. 

See, for example, https://github.com/APS-USAXS/ipython-usaxs/issues/102


prjemian:
Looks right to me, just about to test it now.

prjemian:
Yes, thanks!  That fixes it.

Issue #142, opened 2019-02-19
Race condition in replace_access_rights_event


brunoseivam:
It looks like there is a race condition in `replace_access_rights_event`. It seems that `ctx` is not in `_cache` when this function executes. This is the error I get:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/martins/workspace/bpm/python2.7/linux-x86_64/energy.py", line 238, in build
    return EnergySupport()
  File "/home/martins/workspace/bpm/python2.7/linux-x86_64/energy.py", line 35, in __init__
    for dnum, data in self.bpm_data.items()
  File "/home/martins/workspace/bpm/python2.7/linux-x86_64/energy.py", line 35, in <dictcomp>
    for dnum, data in self.bpm_data.items()
  File "/usr/local/lib/python2.7/dist-packages/epics/pv.py", line 148, in __init__
    callback=self.__on_access_rights_event)
  File "/usr/local/lib/python2.7/dist-packages/epics/ca.py", line 410, in wrapper
    return fcn(*args, **kwds)
  File "/usr/local/lib/python2.7/dist-packages/epics/ca.py", line 959, in replace_access_rights_event
    ch = _cache[ctx][name(chid)]
KeyError: 93833199645328
```

I couldn't reproduce the problem with a simple script (the issue currently happens inside a softioc using pyDevSup). 

I only see the issue when using the pyepics version from PIP. If I use our local version (last upstream commit: [4e45886](https://github.com/pyepics/pyepics/commit/4e458861df1082f881cd4e1ba01b0b4b0e97d8d6)) the issue is not triggered.

@domitto

dchabot:
Hi @brunoseivam.

I made the following patch after stumbling over this in a very similar way (pyepics withing pyDevSup):

```diff
git diff master race_cond_channel_create
diff --git a/epics/ca.py b/epics/ca.py
index 98a34c9..4f7f1c5 100755
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -897,11 +897,11 @@ def create_channel(pvname, connect=False, auto_cb=True, callback=None):
         # already have or waiting on a chid
         chid = _cache[ctx][pvname]['chid']
     else:
-        chid = dbr.chid_t()
+        entry['chid'] = dbr.chid_t()
         ret = libca.ca_create_channel(ctypes.c_char_p(pvn), conncb, 0, 0,
-                                      ctypes.byref(chid))
[...]

PR #143, opened 2019-02-19
FIX: race condition

dchabot:
Addresses issue #142.

Attn: @brunoseivam


newville:
@dchabot Thanks, I don't think I have ever seen that one, but I can see how that could happen.  Kind of surprising that it hadn't been seen, actually. 

Anyone object to merging this ASAP?

Issue #141, opened 2019-02-18
ca.get doesn't work as advertised


brunoseivam:
`ca.get` doesn't seem to work as advertised. The documentation says that a disconnected channel will return None as a value, but it actually throws.

```
python
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from epics import ca
>>> chid = ca.create_channel('dummy')
>>> ca.get(chid, timeout=1.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/epics/ca.py", line 437, in wrapper
    name(chid), timeout))
epics.ca.ChannelAccessException: get() timed out waiting 'dummy' to connect (1 seconds)
>>> ca.get(chid, timeout=1.0, wait=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/epics/ca.py", line 437, in wrapper
    name(chid), timeout))
epics.ca.ChannelAccessException: get() timed out waiting 'dummy' to connect (1 seconds)
```

brunoseivam:
@domitto

newville:
@brunoseivam  Hm, yes I see that too.   At first, one might be inclined to view this as a documentation bug, but I sort of think the "return None" is preferable.   Do you or anyone else have a strong preference?  I doubt we'd be able to please everyone here. 

Maybe the right question is:  Does anyone object to treating this as a behavior bug (and not as documentation bug) and fixing `ca.get()` to return `None` for an unconnected channel?   

FWIW, I think the solution is just a matter of replacing the `withConnectedCHID` decorator with a `withCHID` decorator.

danielballan:
Since you ask, I'll share our experience with this question in caproto. We experimented with always raising `TimeoutError` on timeout, never returning `None`. Anecdotally, in our use of caproto on the experimental floor at BNL so far, we have found raising to be beneficial. It tends to produce clearer error messages in common usages and does not remove any flexibility.

[...]

PR #144, opened 2019-02-21
allow get to return  (as documented) for unconnected Channels

newville:
This allows `ca.get()` and related `ca.get***` to run with an unconnected channel ID and return `None` for an unconnected channel,  and not simply raise an exception in the decorator.  

This should address #141

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -413,6 +413,7 @@ def withCHID(fcn):
     wrapper.__dict__.update(fcn.__dict__)
     return wrapper
 
+
 def withConnectedCHID(fcn):
     """decorator to ensure that the first argument of a function is a
     fully connected Channel ID, ``chid``.  This test is (intended to be)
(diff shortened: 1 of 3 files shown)


brunoseivam:
Hi @newville, the current state of this branch seems to work for me. There's one wrinkle, though: it seems that `timeout` is not being respected:

```
python -m timeit -n 3 -s 'import epics; pvs = ["dummy%d" % i for i in range(3)]' 'epics.caget_many(pvs, timeout=1.0)'
3 loops, best of 3: 24 sec per loop
```

newville:
@brunoseivam Yes, `timeout` for `caget_many()` is not very specific.  We don't know really apply that as a global timeout, and the default connection timeout (2 seconds) may be used many times within `caget_many()`.   To be clear, `caget_many()` does:
  1. create many channels without callbacks (should be very fast)
  2. try to connect to many  channels (may need to wait for connections)
  3. ask for "gets" on all channels without waiting for result (will try to connect if not connected)
  4. ask for the gets to compete (should be fast)

I'm sure this could be improved.... suggestions or PR welcome!

brunoseivam:
Would you be willing to accept something like the code in [this comment](https://github.com/pyepics/pyepics/issues/141#issuecomment-465162177)? I don't know if `ca.poll()` can be called in a tight loop like this. I can open a PR

newville:
@brunoseivam Yeah, I think that would be a good start to a `caget_many` with a more deterministic timeout.  I think it is completely fine to use `ca.poll` in a loop like you have it.

brunoseivam:
I think this can be closed, then. I will open a PR next week with the changes to `caget_many`. Thanks!

newville:
@brunoseivam  OK, I'll merge this now and look forward to a PR for a `caget_many` that better respects the `timeout` argument.

klauer:
Taking a look at master now, seems like `raise_on_unconnected` from this PR is not defined in `ca.py`. Was the intention to make `withConnectedCHID` a configurable wrapper - optionally raising?

newville:
@klauer Oops, yes it looks like that `or raise_on_unconnected` should be removed.  I don't remember what that was supposed to do ;) 

We must not be testing that very carefully, perhaps because `withConnectedCHID` is now only used for `ca.put()` and none of the tests try to do a `put()` on an unconnected chid?

I'd be inclined to just push a fix to this to master within a few days.

PR #147, opened 2019-04-06
REF/FIX: cache race conditions, put_complete, and more...

klauer:
(This PR is on the larger side - I can break it up if some parts are more controversial than others)

Added tests
-------------
This PR adds two tests that fail on master:
  1. ~pv_unittest.test_multithreaded_get
    This fails on master due to race conditions with the per-context cache~ pulled out into #150 
  2. ~pv_unittest.test_multithreaded_put_complete
    This fails on master since put_completes are keyed solely on `pvname` and there is no guarantee that a libca callback is corresponding to the specific user request~ pulled out into #149 

Fixes as implemented in this PR
--------------------------------

* ~Fix for (2): Changing the `put_complete` callbacks to be unique for each request - clients can now be sure that the respective call was actually completed~ pulled out into #149 
* Fix for (1): Making it such that simultaneous get requests can share the result. In master, this is a race to "who gets the result first", with the rest failing somewhere in `get_complete` (or receiving `None`)
    * The logic is now as follows: 
        * If there is an in-process get request for a specific field type (e.g., DBR_CTRL_INT) for a PV, other callers to `get()` with those same parameters will share the result and not duplicate the `libca` call. 
    * `get()` and `get_complete()` are two separate calls in pyepics. That means that in-between a call to `get()` and `get_complete()` on the user-side, it's possible that one or more duplicate get requests may have happened in separate threads. The end result would be that each call to `get_complete()` would get the latest value, and not necessarily the one that was in-flight when `get()` itself was called. I don't see an obvious way around this with the current API, but I don't necessarily think it's much of an issue, either.
    * There is a (mostly) minimal set of locking in place to keep track of these requests. 
    * Now, the codebase is more rigorous about the per-context cache bookkeeping
        * Adds a `_CacheItem` class that handles some of this. It keeps back-compat with `__getitem__` - but really clients should _not_ be touching the cache at all.

Additional issue and fix
---------------------------
`ca.get()` can fail and not return _any_ result to the user, ending in a timeout condition. This happens due to the check in `_onGetEvent`: https://github.com/pyepics/pyepics/blob/7fe180e0ab857f8189b4e1b048608235ac45b045/epics/ca.py#L613

This is resolved in this PR with the exception status propagating back to the user. This is a slight API break - errors that used to be masked as timeouts will now raise `ChannelAccessGetFailure`.

Miscellaneous fixes
--------------------
* ~Use `functools.wraps()` where needed~ - pulled out into #148 
* Tweak `GET_PENDING` to have a nice repr
* Use `defaultdict` to remove a bunch of `if ca.current_context() not in _cache` checks
* Access rights, connection callbacks were being run for all contexts
    * That did not make sense to me, and it is not present here. If I misunderstood that, we should revisit this.
* Refactors `get_timevars`, `get_ctrlvars` using `get_with_metadata`. All `get_*()`-related functions are now routed through this function only.

Overall
--------
This is a rather big change under the hood and should be reviewed/tested carefully. The implementation details are all negotiable.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -411,24 +411,6 @@ def get_cache(pvname):
     return _cache[current_context()].get(pvname, None)
 
 
-def _matching_cache_entries(pvname, chid=None):
-    "return all matching cache entries for a specific pvname, chid"
-    # TODO: This is as previously implemented - I'm not sure if it actually
-    # makes sense. Even if pvnames/chids are the same from different contexts,
-    # they are not _actually_ the same.
-    matching_pvnames = [
-        (context, cvals[pvname])
-        for context, cvals in _cache.items()
-        if pvname in cvals
-    ]
-    if chid is None:
-        return matching_pvnames
-    return [(context, entry)
-            for context, entry in matching_pvnames
-            if entry.chid == chid
-            ]
-
-
 def show_cache(print_out=True):
     """print out a listing of PVs in the current session to
     standard output.  Use the *print_out=False* option to be
@@ -657,27 +639,19 @@ def _onConnectionEvent(args):
     if dbr.PY64_WINDOWS:
         args = args.contents
 
-    ctx = current_context()
     conn = (args.op == dbr.OP_CONN_UP)
-
-    if ctx is None and len(_cache.keys()) > 0:
-        ctx = list(_cache.keys())[0]
 
     # search for PV in any context...
     pvname = name(args.chid)
-    chid = int(args.chid)
-    entries = _matching_cache_entries(pvname, None)
+    entry = get_cache(pvname)
 
     # logging.debug("ConnectionEvent %s/%i/%i " % (pvname, args.chid, conn))
     # print("ConnectionEvent %s/%i/%i " % (pvname, args.chid, conn))
-    if not entries:
-        entry = _CacheItem(chid=chid, pvname=pvname)
-        _cache[ctx][pvname] = entry
-        entries = [(ctx, entry)]
+    if entry is None:
+        entry = _CacheItem(chid=args.chid, pvname=pvname)
+        _cache[current_context()][pvname] = entry
 
-    # set connection time and run connection callbacks in all contexts
-    for context, entry in entries:
-        entry.run_connection_callbacks(conn=conn, timestamp=timestamp)
+    entry.run_connection_callbacks(conn=conn, timestamp=timestamp)
 
 
 ## get event handler:
@@ -735,8 +709,10 @@ def _onAccessRightsEvent(args):
 
     # Getting bunk result from ca.current_context on channel disconnect
     # Do this the long way...
-    for context, entry in _matching_cache_entries(pvname, chid):
+    entry = get_cache(pvname)
+    if entry is not None:
         entry.run_access_event_callbacks(ra, wa)
+
 
 # create global reference to these callbacks
 _CB_CONNECT = dbr.make_callback(_onConnectionEvent, dbr.connection_args)
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -15,14 +15,16 @@ documentation here is developer documentation.
 import ctypes
 import ctypes.util
 
+import atexit
+import collections
 import functools
 import os
 import sys
+import threading
 import time
-from math import log10
-import atexit
 import warnings
-from threading import Thread
+
+from math import log10
 from pkg_resources import resource_filename
 
 from .utils import (STR2BYTES, BYTES2STR, NULLCHAR, NULLCHAR_2,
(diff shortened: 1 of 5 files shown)


newville:
@klauer OK, that's a lot of changes!  I'm not entirely sure I understand the extent of the issues this addresses, so it might take me some time to come up to speed on this.

If I understand there are two main issues to be solved, both related to multi-threading, and then also a fair amount of other changes that are "miscellaneous".   

The biggest code change appears to be a complete refactoring of the ca-level cache to solve a problem with multi-threaded gets.  What is that problem?  Is it that threads A and B  do overlapping  `ca.get(wait=False)` and only one `ca.get_complete()` works correctly?

Is the problem with multi-threaded put-with-complete similar?  If not, what is that problem?

I can certainly believe that there are race condition problems with really using simultaneous calls to the same PVs from multiple threads.  I don't think we're testing that very hard . Anyway, while I'm not opposed to any of these changes, I am slightly concerned about unintended consequences.

It would be clearer (to me, anyway!) if these were separate issues and PRs, especially as they appear to overlap in changes to `ca.py`.  I think I'm OK with the cleanups that you label as 'miscellaneous', but again, it might be better to do these in sequence.

OTOH, I'm also completely willing to say that you understand threading and low-level CA much better than me and so basically defer to you on all of this.

klauer:
@newville - that's all fair. I'll try to break this up into smaller, more easily understandable chunks. 

That'll take a fair amount of time to make sure I do it right, so I'll just leave this open until I get a chance to do that.

newville:
@klauer I feel like I should explain my caution here.  For other projects, I'm much more willing to make and accept large changes because I fundamentally trust the tests.   For the "simple parts" of pyepics, I think this works too.  But it also seems like many of the recent issues and changes have really been about how pyepics wraps CA, especially dealing with what CA does at a fairly low level, or what *I* would normally consider "advanced topics"  such as connection callbacks,  using multiple threading, etc.   I am not sure that we really have solid tests for these. I am certainly not an expert in threading or network protocol but I think it is hard to have automated tests for "when the network is a bit slow" or multi-threaded uses with an imperfect network.  Somehow, I think this is more than just "oh we didn't have a test for that".

I've also heard rather too many rude and frankly idiotic complaints about how pyepics works or is structured from people in the epics community -- complaints about pyepics causing DDOS, seemingly endless complaints about the distributed libca.so on Linux, complaints about having to use PYEPICS_LIBCA, or insisting that installing with NO_LIBCA ever makes any sense.   I support many software projects and try to help as many people as I can with many different projects.  I have to admit that I have never been less inclined to support other people using code that I actually use.

So, taking over the project is certainly an option.   If you aren't willing to do that, please raise individual issues and solve them one at a time.

klauer:
@newville, you have every right to be cautious! I can certainly understand your frustration.  

Regardless of the implementation details, pyepics provides a very convenient and useful method to access EPICS in Python. People should be grateful for that alone - and make PRs, working with us to improve it for everyone - rather than complain about everything. I'm trying to do my part, in any case. 

For this PR, some of the minor fixes can be split out at the very least. Perhaps we can also work on making some more rigorous threading tests.

newville:
@klauer Thanks, and to be clear none of my frustration is with you!  

Would it be possible and sensible to make one PR that improves the CA cache to a class and then another (or maybe 2?) that then re-do the "get-with-callback" and "put-with-callback"?   When reading the current changes, I cannot always tell how those are broken up.    I'm definitely OK with improving the CA cache.

The "get-with-callback" seems like a pretty unusual thing to me.  So, I can believe there are problems.  If two threads to overlapping "get-with-callback", they should both succeed, even if there is no guarantee that the values would be the same.

OTOH, "put-with-callback" seems completely normal to me. But in the case of two different threads doing this at the same time, I'm not sure I know what correct behavior is supposed to be. Let's say PV1 currently as value 0. Thread (or process) A puts PV1 to 10,  which might take 10 seconds to complete.  A second later Thread (or process) B puts PV1 to 3.     Does the PV first go to10 and *then* go to 3?   I think that might depend on the details of the record type (like, for a Motor it probably will *not* go to 10).  That is, I don't think CA guarantees that it will.  I think a "put-with-callback" has to be able to fail gracefully with "did not complete", and it seems like we have a choice for what Thread A does in this case.

FWIW, it is also OK with me if "multi-thread" and "multi-process" give different results, especially when using "attach to original thread context".

klauer:
That seems reasonable. I can definitely break out a few things here and will do so.

Yeah, the primary concern of the `get/get_complete` is more that it will fail miserably under the right conditions. I believe what's in this PR is closer to "correct", or at least what one might expect given the race condition.

For the put-with-callback, the exact response of the IOC is not our (library) concern. It's important to split up the client and server's role in this case. The logic is something like the following for a single request: "client request A was made. Did request A finish according to the IOC?" The current master has no issues with that scenario.  For multiple requests, whether from different threads or even at different points in the same function: "client requests A and B were made. Did request A finish? Did request B finish?" In the current master, it's impossible to tell if the put completion notification was for request A or request B. This PR makes an attempt at fixing that.

klauer:
OK, I did what I could to separate these. No rush on the review, @newville - but hopefully it makes the process even slightly easier.

PR #148, opened 2019-04-08
REF: use functools.wraps

klauer:
Current master manually wraps functions, but there is a utility available in the standard library to do this for you. Verified it's available in 2.7 and up:
https://docs.python.org/2/library/functools.html#functools.wraps


newville:
@klauer  thanks! looks simple, and a definite improvement. Let's just do this one.

PR #149, opened 2019-04-08
REF: Add and fix test pv_unittest.test_multithreaded_put_complete

klauer:
Adds a test `pv_unittest.test_multithreaded_put_complete` which fails on master since put_completes are keyed solely on pvname (irrespective of context or anything else). There is no guarantee that the callback received corresponds to any specific user request.

This issue could be potentially seen in single-threaded cases, but the included test is done with many threads to provide some degree of certainty that it's doing what is expected.

My notes from the other PR this was separated from:
> For the put-with-callback, the exact response of the IOC is not our (library) concern. It's important to split up the client and server's role in this case. The logic is something like the following for a single request: "client request A was made. Did request A finish according to the IOC?" The current master has no issues with that scenario. For multiple requests, whether from different threads or even at different points in the same function: "client requests A and B were made. Did request A finish? Did request B finish?" In the current master, it's impossible to tell if the put completion notification was for request A or request B. This PR makes an attempt at fixing that.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -82,6 +82,7 @@ DEFAULT_CONNECTION_TIMEOUT = 2.0
 #                  change (accumulated in the cache)
 _cache  = {}
 _namecache = {}
+_put_completes = []
 
 # logging.basicConfig(filename='ca.log',level=logging.DEBUG)
 ## Cache of pvs waiting for put to be done.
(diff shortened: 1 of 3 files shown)


newville:
@klauer This looks good and I think we can merge this.  Do you agree that `PV._put_complete` should be initialized,  either to `True` or perhaps `None` to signal "not even started a put-with-callback"?

newville:
This looks good to me - Thanks!

PR #150, opened 2019-04-08
REF: rework CA cache for get requests (improved thread safety and such)

klauer:
This is the largest change from #147 that couldn't be easily separated. It's also the part that deserves the most scrutiny prior to (potential) merging.

Issues
-------
1. Multiple threads attempting to read the same PV quickly fail due to a race condition with the built-in `epics.ca._cache`
2. `ca.get()` which returns anything but `ECA_NORMAL` will show up in pyepics as `TimeoutError`
3. `get_timevars`, `get_ctrlvars` duplicate a lot of `get_with_metadata` code unnecessarily.
4. Access rights, connection callbacks were being run for all contexts
    This did not make sense to me, and it is not present in this refactor. If I misunderstood that, we should revisit this. (@dchabot, I believe this was from your PR?)
5. `_namecache` is not keyed on context. This PR removes the namecache entirely.

Fix to issue 1
-------------
Added test `pv_unittest.test_multithreaded_get` shows how to easily trigger this failure.

 In master, this is a race to "who gets the result first", with the rest failing somewhere in `get_complete` (or receiving `None`). With this PR, simultaneous get requests can now share the result.
The logic is now as follows: 
   * If there is an in-process get request for a specific field type (e.g., DBR_CTRL_INT) for a PV, other callers to `get()` with those same parameters will share the result and not duplicate the `libca` call. 
   * `get()` and `get_complete()` are two separate calls in pyepics. That means that in-between a call to `get()` and `get_complete()` on the user-side, it's possible that one or more duplicate get requests may have happened in separate threads. The end result would be that each call to `get_complete()` would get the latest value, and not necessarily the one that was in-flight when `get()` itself was called. I don't see an obvious way around this with the current API, but I don't necessarily think it's much of an issue, either.
  * There is a (mostly) minimal set of locking in place to keep track of these requests. 
  * Now, the codebase is more rigorous about the per-context cache bookkeeping
      * Adds a `_CacheItem` class that handles some of this. It keeps back-compat with `__getitem__` - but really clients should _not_ be touching the cache at all.

Fix to issue 2
------------------------
`ca.get()` can fail and not return _any_ result to the user, ending in a timeout condition. This happens due to the check in `_onGetEvent`: https://github.com/pyepics/pyepics/blob/7fe180e0ab857f8189b4e1b048608235ac45b045/epics/ca.py#L613

This is resolved in this PR with the exception status propagating back to the user. This is a slight API break - errors that used to be masked as timeouts will now raise `ChannelAccessGetFailure`.

Fix to issue 3
--------------
* Refactors `get_timevars`, `get_ctrlvars` using `get_with_metadata`. All `get_*()`-related functions are now routed through this function only.

Fix to issue 4
-------------
* Only the access rights/connection callbacks for the specific context are run after this PR.

Miscellaneous improvements
-----------------------------
* Tweak `GET_PENDING` to have a nice repr
* Use `defaultdict` to remove a bunch of `if ca.current_context() not in _cache` checks

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -392,24 +392,6 @@ def get_cache(pvname):
     return _cache[current_context()].get(pvname, None)
 
 
-def _matching_cache_entries(pvname, chid=None):
-    "return all matching cache entries for a specific pvname, chid"
-    # TODO: This is as previously implemented - I'm not sure if it actually
-    # makes sense. Even if pvnames/chids are the same from different contexts,
-    # they are not _actually_ the same.
-    matching_pvnames = [
-        (context, cvals[pvname])
-        for context, cvals in _cache.items()
-        if pvname in cvals
-    ]
-    if chid is None:
-        return matching_pvnames
-    return [(context, entry)
-            for context, entry in matching_pvnames
-            if entry.chid == chid
-            ]
-
-
 def show_cache(print_out=True):
     """print out a listing of PVs in the current session to
     standard output.  Use the *print_out=False* option to be
@@ -639,27 +621,19 @@ def _onConnectionEvent(args):
     if dbr.PY64_WINDOWS:
         args = args.contents
 
-    ctx = current_context()
     conn = (args.op == dbr.OP_CONN_UP)
-
-    if ctx is None and len(_cache.keys()) > 0:
-        ctx = list(_cache.keys())[0]
 
     # search for PV in any context...
     pvname = name(args.chid)
-    chid = int(args.chid)
-    entries = _matching_cache_entries(pvname, None)
+    entry = get_cache(pvname)
 
     # logging.debug("ConnectionEvent %s/%i/%i " % (pvname, args.chid, conn))
     # print("ConnectionEvent %s/%i/%i " % (pvname, args.chid, conn))
-    if not entries:
-        entry = _CacheItem(chid=chid, pvname=pvname)
-        _cache[ctx][pvname] = entry
-        entries = [(ctx, entry)]
+    if entry is None:
+        entry = _CacheItem(chid=args.chid, pvname=pvname)
+        _cache[current_context()][pvname] = entry
 
-    # set connection time and run connection callbacks in all contexts
-    for context, entry in entries:
-        entry.run_connection_callbacks(conn=conn, timestamp=timestamp)
+    entry.run_connection_callbacks(conn=conn, timestamp=timestamp)
 
 
 ## get event handler:
@@ -724,8 +698,10 @@ def _onAccessRightsEvent(args):
 
     # Getting bunk result from ca.current_context on channel disconnect
     # Do this the long way...
-    for context, entry in _matching_cache_entries(pvname, chid):
+    entry = get_cache(pvname)
+    if entry is not None:
         entry.run_access_event_callbacks(ra, wa)
+
 
 # create global reference to these callbacks
 _CB_CONNECT = dbr.make_callback(_onConnectionEvent, dbr.connection_args)
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -15,14 +15,17 @@ documentation here is developer documentation.
 import ctypes
 import ctypes.util
 
+import atexit
+import collections
 import functools
+import logging
 import os
 import sys
+import threading
 import time
-from math import log10
-import atexit
 import warnings
-from threading import Thread
+
+from math import log10
 from pkg_resources import resource_filename
 
 from .utils import (STR2BYTES, BYTES2STR, NULLCHAR, NULLCHAR_2,
@@ -74,13 +77,8 @@ AUTOMONITOR_MAXLENGTH = 65536 # 16384
 DEFAULT_CONNECTION_TIMEOUT = 2.0
 
 ## Cache of existing channel IDs:
-#  pvname: {'chid':chid, 'conn': isConnected,
-#           'ts': ts_conn, 'callbacks': [ user_callback... ])
-#  isConnected   = True/False: if connected.
-#  ts_conn       = ts of last connection event or failed attempt.
-#  user_callback = one or more user functions to be called on
-#                  change (accumulated in the cache)
-_cache  = {}
+# Keyed on context, then on pv name (e.g., _cache[ctx][pvname])
+_cache = collections.defaultdict(dict)
 _namecache = {}
 
 # logging.basicConfig(filename='ca.log',level=logging.DEBUG)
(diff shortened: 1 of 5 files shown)


klauer:
@dchabot this callback happens below the `epics.PV` object-level - for one specific chid from one specific context.

Old - look in all contexts for a specific PV name:
```python
    for ctx in _cache.values():
        pvname = name(chid)
        if pvname in ctx:
            ch = ctx[pvname]
            if 'access_event_callback' in ch:
                for callback in ch['access_event_callback']:
                    if callable(callback):
                        callback(ra, wa)
```

This will give multiple callbacks if several contexts access the same PV name (different context, different channel id).

New - `get_cache` assumes `current_context()` and only gives you that instance:
```python
    entry = get_cache(pvname)
    if entry is not None:
        entry.run_access_event_callbacks(ra, wa)
```

Hmm, I may have skipped over this comment in refactoring, though:
```python
    # Getting bunk result from ca.current_context on channel disconnect
    # Do this the long way...
```

Is this something that's really a concern? If the channel is disconnected, access rights don't really mean much, I'd think.

klauer:
@dchabot OK, I can see your train of thought there. Double-checking the docs to confirm (https://epics.anl.gov/base/R3-14/10-docs/CAref.html#ca_replace) shows that clients get access rights callbacks per-chid (in one context). Since individual `epics.PV`s always have a unique `(ctx, chid)` pair, I don't think there's any further to be concerned for this specific issue.

As to the `KeyError`s that you saw, except maybe if there was a race between the access rights callback and the connection callback (... ?), I can't imagine where those came from.

If you have a chance to give this PR a try, mind letting me know if you see any problems?

newville:
@dchabot @klauer FWIW, I would suspect that if  Channel ID to Channel Name was not exactly one-to-one within a context (and for a DBR type) that many applications would break.  I think the C interface really uses CHID as the pointer and Name as a property of that pointer.    If a client app runs out of c_long pointers to Channels, I think many problems would happen.

Out of curiosity are either of you (or anyone else) using multiple threads and *not* using the initial context in each non-main thread?   If no one is actually using per-thread contexts, should we consider saying that we don't support anything except attaching to the initial context and then not include the context as part of the cache key?  I do not have a strong opinion either way: per-thread contexts seem to work in simple cases, but the bookkeeping has some cost.

klauer:
I think there's some benefit to retaining the ability to use multiple contexts. Perhaps, though, we could modify the default behavior: unless the user specifically requests a new context in a thread, use the initial context. That would obviate the need for `epics.ca.CAThread`.

As to the `KeyError`, assuming you don't share `epics.PV` objects between threads in different contexts, after this PR I think it should not occur. It may also be possible to detect when the user is erroneously doing something like this and raise a more appropriate exception.

I'll have to look into the above 2 thing. I think they would be on top of this PR (and not conflicting).

klauer:
Another context-related issue: `namecache` is keyed only on `chid` as of now.

Thinking about it a bit more, the minor performance benefit of the cache does not seem to be worth the additional bookkeeping complexity of `namecache`, and would suggest its removal entirely (end result being: `def name(chid) -> libca.name(chid)`). @newville, do you have any strong feelings either way?

newville:
@klauer  The name cache does seem unnecessary.  Keeping Context as part of the cache key is fine with me. I guess the simplest use of threads without attaching to the initial context or using CAThread will work most of the time.  I guess the question is how much guidance to give the user to use the initial context.  Anyway, I'm happy to defer to you all on this.   I think you are more in touch with what the people *needing* threads (perhaps the PyDM folks?) might expect.

klauer:
@newville OK, I will see what I can come up with, threading-wise.

Since this PR goal is to refactor the cache, I removed the namecache (as we discussed above) in 25914a5 instead of putting it in another PR.

archman:
> @klauer The name cache does seem unnecessary. Keeping Context as part of the cache key is fine with me. I guess the simplest use of threads without attaching to the initial context or using CAThread will work most of the time. I guess the question is how much guidance to give the user to use the initial context. Anyway, I'm happy to defer to you all on this. I think you are more in touch with what the people _needing_ threads (perhaps the PyDM folks?) might expect.

When I'm developing high-level apps with PyQt and pyepics, for the modularity, the PV objects are passing between different UI widgets, or even different apps, that is they actually spreading in different threads, and normally the race condition happens, at least for Python 3.5, but Python 3.6 (3.7) can work without any issues. The interesting thing @dchabot and me found is if running the app with `%pdb on` in IPython console, the issue is gone, seems like the execution timing is changed in pdb mode, which we guess Python 3.6 and later releases must has done something at this point.

FWIW, the PyDM widget normally can work without any issue, since the PV object is created and used with the widget, if it would be passed back and forth, the similar issue might happen.

hhslepicka:
@archman for PyDM it works because no widget has control over the PV and many widgets share the same PV having its data being transfered over Qt Signals and Slots which makes the conflicts of threads go away.

archman:
> @archman for PyDM it works because no widget has control over the PV and many widgets share the same PV having its data being transfered over Qt Signals and Slots which makes the conflicts of threads go away.

@hhslepicka I assume the PV you mentioned is PV object? And the PV object is being transferred over Qt signal to slots?

hhslepicka:
No PV object is transferred ever. The PV object for PyDM belongs to the Connection at the Data Plugin that is shared between widgets. What gets transferred is the data and metadata from a PV.

newville:
@archman 
> When I'm developing high-level apps with PyQt and pyepics, for the modularity, the PV objects are passing between different UI widgets, or even different apps, that is they actually spreading in different threads, and normally the race condition happens, at least for Python 3.5, but Python 3.6 (3.7) can work without any issues. 

Sorry, but I do not understand what you are saying. What does "the PV objects are passing between different UI widgets" mean?  

For sure, passing PV objects between different threads would be a risky thing to do.  Please read  http://pyepics.github.io/pyepics/advanced.html#how-to-work-with-ca-and-threads and please to do not opt for option 5 ("ignore this advise and hope for the best").  It is much better to 

I have no idea how you could be transferring PV objects between different apps.  Are you trying to use Pickle or something?  I don't think that would work.

So:  what are you doing?   Please provide real code.  Until I read your message here, I did not even know you were using PyQt.  There are lots of examples of GUIs using Pyepics.  I have 4 or 5 GUIs that use pyepics running full time at my beamline.   

> The interesting thing @dchabot and me found is if running the app with %pdb on in IPython console, the issue is gone, seems like the execution timing is changed in pdb mode, which we guess Python 3.6 and later releases must has done something at this point. 

Again: if this happens only with Python3.5 but not with Python 3.6 or Python 3.7, isn't the solution sort of obvious?

> FWIW, the PyDM widget normally can work without any issue, since the PV object is created and used with the widget, if it would be passed back and forth, the similar issue might happen.

What does "passed back and forth" mean to you?  

With wxPython, I have subclassed many widgets to have a PV associated with them.  Thus I can create a PVTextControl that the user can input a value and that automatically gets put() to the PV.  If the PV value updates, the displayed value is automatically updated.  This *does* require some care when definint the wx-epics widgets so that the two event loops do not interfere with one another.  But it is not that hard, and it does not actually use extra Threads.   It just uses the wxPython way to say "handle any wx events generated in this Epics callback function after this Epics code is run".  I believe that is actually similar to simply emitting a Qt Signal to cause the eventual update of the appropriate widgets. 

It is not clear to me that your issues are related to this PR.  So, sorry @klauer.

newville:
@klauer This looks very good to me.  I *think* all the changes are correct, and only a few of them terrify me ;).  FWIW, the Windows-64 bits in pyepics always terrifies me.

Are there any outstanding concerns except "this is a large change to deepest and most-dragon-infested parts of pyepics"?  

Anyway, I'm OK with merging this, especially if we're willing to leave a window open (perhaps say until early to mid May) to allow for the possibility of additional fixes before releasing.  I would suggest calling this 3.4.0, as it appears to be a large implementation change addressing (or at least trying to address) some long-standing issues.

Any objections to that approach?

klauer:
@newville Great! I'm glad you think so. 👍 Everything in scope of "fix the CA cache" is done as far as I'm concerned, though I'll of course continue to build on this work in other PRs.

I don't believe there are any major gotchas/concerns here. One minor one: I did not get around to updating the docs regarding the API change of issue (2).

newville:
@klauer  OK, I think there are no objections, and all existing tests are passing. 

So, I will merge this and tag as 3.4.0rc1.

klauer:
Glad to hear it, @archman! Please let us know in an issue if you run into anything further using master. I'll try to resolve anything you might find prior to the next official release.

newville:
@archman  it would be very helpful if you would provide actual code that showed the problem you were having.  That way, we could understand what you are doing and have the option to add a test for it so that we ensure it works.  As it is, I have no idea what you are talking about.    Of course

     from epics import caget; print(caput(<some-pv-name>, <value>))

will not work unless `caput` had been previously imported, but it won't give a connection event.  That simply can not be the error message you actually got from the code you posted.  

What does "I tried to do something after IocInit()" mean to you?  

I've asked you for example code a number of times now, and you have been remarkably unwilling to help us understand the problems you are seeing.   I think you may not be understanding some fundamental aspects of how pyepics works, and I am certain you are not participating in conversations about pyepics that is helpful to anyone else.

archman:
@newville
> will not work unless caput had been previously imported,

I'm sorry, that was a typo, should be ` from epics import caput`. 

> I've asked you for example code a number of times now, and you have been remarkably unwilling to help us understand the problems you are seeing. I think you may not be understanding some fundamental aspects of how pyepics works, and I am certain you are not participating in conversations about pyepics that is helpful to anyone else.

And the issue I'm talking about this time has nothing to do with GUI (for that case I cannot provide a simple enough script here). I'm happy to provide any example I can, below is the `st.cmd` file for testing (to support `py` lines, you need to have `pydevsup` installed):
```
#!../../bin/linux-x86_64/ems

< envPaths

epicsEnvSet("PYEPICS_LIBCA", "$(EPICS_BASE)/lib/linux-x86_64/libca.so")
epicsEnvSet("PREFIX", "EMS")

dbLoadDatabase("${TOP}/dbd/ems.dbd",0,0)
ems_registerRecordDeviceDriver(pdbbase)

dbLoadRecords("$(TOP)/db/ems.db", "P=$(PREFIX)")

iocInit()

py "import post_init"
```

And the `post_init.py` could be anything that you'd like to do with the IOC, e.g.:
```
from epics import caput

pv = 'pos_stop'
val = 10.0
caput(pv, val)
```

> What does "I tried to do something after IocInit()" mean to you?

See the line startswith `py` after `iocInit()` line.

Again, this is merely my use experience with this package, sometimes it is not practical to present all of the code as an example, though I'm trying my best to construct a simple enough demo to show the issue. Maybe I'm not getting all your design idea about this package, maybe I'm using the pacakge in the wrong way, maybe people normally do not do post-initialization like my way, still there might be chances that some people encounter this issue, and hopefully they could be directed here by googling the similar error message, and maybe a simple hint like 'just try pyepics with this PR' may solve their problem.

newville:
@archman   Well, this (or any other) closed PR is a poor choice of a place to have such a discussion.  It is not at all clear to me that your problem is related to this PR.  If you have trouble with pyepics, open a new issue, ask a question, and provide example code and full traceback. 

Should we expect that pyepics works with pydevsup?  Honestly, I do not know.  I'm not saying we should not support it, but we currently actually do not.  Specifically, we do not have any tests for using pyepics with pydevsup.   Perhaps pydevsup has tests using pyepics?  Dunno.   Maybe it will "just work".  Then again, I would not be surprised if there were severe restrictions on what the Python interpreter embedded into the IOC could and could not do and I can certainly believe that there could be issues related to networking, threading, and resource management.

Issue #145, opened 2019-03-04
No monitor for large PVs after clear_auto_monitor()


hinxx:
After creating a ```pv = epics.PV()``` with ```auto_monitor=True``` I get the value updates as expected (the value callback gets called). If I need to stop the callbacks (monitor) I call ```clear_auto_monitor()``` and later ```reconnect()``` to get the updates going again. All good so far.

The above procedure fails for me when large arrays (~500k points; ~1MB) are at stake.
The last step - reconnect() - does not seem to re-enable the value change updates to call my callback anymore. No other warnings or errors were noticed  so far..

Any ideas why this manifests itself for larger arrays only, while it works for scalars and smaller arrays (ie. ~900 points)?

dchabot:
Hi @hinxx.

I just had a quick look, so maybe mistaken, but it looks like there's a bug in `reconnect()` -- it resets `PV.auto_monitor = None`, whereas you've initialized the `epics.PV` with `auto_monitor=True`.

`PV.auto_monitor = None` has the side-effect of omitting the channel subscription (i.e. monitor for changes) if `count < ca.AUTOMONITOR_MAXLENGTH` (iirc, equivalent to the default EPICS_CA_MAX_ARRAY_BYTES of 16k).

A quick check of this theory would be to see if transfers succeed after a `clear_auto_monitor()` --> `reconnect()` cycle for payloads < 16kB, but fail for payload > 16kB.

newville:
@hinxx @dchabot  Oh, very sorry about that.  I agree that this clearing auto_monitor then reconnecting looks to be broken for large arrays that are monitored.   

It maybe that either `reconnect()` or `force_connect()` (or maybe both) needs an `auto_monitor` keyword arg (with default of `None` meaning the usual "automonitor if not too large")...  Does that seem right to you?

dchabot:
> @hinxx @dchabot Oh, very sorry about that. I agree that this clearing auto_monitor then reconnecting looks to be broken for large arrays that are monitored.
> 
> It maybe that either `reconnect()` or `force_connect()` (or maybe both) needs an `auto_monitor` keyword arg (with default of `None` meaning the usual "automonitor if not too large")... Does that seem right to you?

Yes, I think _both_ `reconnect()` and `force_connect()` should get a `auto_monitor` kwarg.
[...]

PR #151, opened 2019-04-16
ENH: add _ensure_context decorator, guard against multithreaded access of `epics.PV`

klauer:
Allows any thread, using any CA context, to access an `epics.PV` object safely.

The modified test verifies the functionality alongside `epics.CAThread`, which is largely made unnecessary with this change.

A much simpler example of which this PR aims to fix, which will fail on master (click for details):

<details>

```python
import threading
import epics


def test(i):
    print()
    print('---- test {} ----'.format(i))
    epics.ca.create_context()
    print('created context', epics.ca.current_context())

    print(pv.get())
    print(pv.nelm)
    print(epics.ca._cache)


pv = epics.PV('Py:ao1', auto_monitor=False)
pv.wait_for_connection()

for i in range(5):
    thread = threading.Thread(target=test, args=(i, ))
    thread.start()
    thread.join()
```

With the current master, we see:
```sh
$ python threaded_access.py
---- test 0 ----
created context 140197677284720
None
1
defaultdict(<class 'dict'>, {140197715687376: {'Py:ao1': <epics.ca._CacheItem object at 0x105987860>}, 140197677284720: {}})

---- test 1 ----
created context 140197679474720
None
1
defaultdict(<class 'dict'>, {140197715687376: {'Py:ao1': <epics.ca._CacheItem object at 0x105987860>}, 140197677284720: {}, 140197679474720: {}})

---- test 2 ----
created context 140197677388512
None
1
defaultdict(<class 'dict'>, {140197715687376: {'Py:ao1': <epics.ca._CacheItem object at 0x105987860>}, 140197677284720: {}, 140197679474720: {}, 140197677388512: {}})

---- test 3 ----
created context 140197679546976
None
1
defaultdict(<class 'dict'>, {140197715687376: {'Py:ao1': <epics.ca._CacheItem object at 0x105987860>}, 140197677284720: {}, 140197679474720: {}, 140197677388512: {}, 140197679546976: {}})

---- test 4 ----
created context 140197677426480
None
1
defaultdict(<class 'dict'>, {140197715687376: {'Py:ao1': <epics.ca._CacheItem object at 0x105987860>}, 140197677284720: {}, 140197679474720: {}, 140197677388512: {}, 140197679546976: {}, 140197677426480: {}})
```

With this PR, we see:

```sh
$ python threaded_access.py

---- test 0 ----
created context 140232599604656
5.33379434247252
1
defaultdict(<class 'dict'>, {140232600730528: {'Py:ao1': <epics.ca._CacheItem object at 0x10860db70>}})

---- test 1 ----
created context 140232600755584
5.33379434247252
1
defaultdict(<class 'dict'>, {140232600730528: {'Py:ao1': <epics.ca._CacheItem object at 0x10860db70>}})

---- test 2 ----
created context 140232638468048
5.33379434247252
1
defaultdict(<class 'dict'>, {140232600730528: {'Py:ao1': <epics.ca._CacheItem object at 0x10860db70>}})

---- test 3 ----
created context 140232601293600
5.33379434247252
1
defaultdict(<class 'dict'>, {140232600730528: {'Py:ao1': <epics.ca._CacheItem object at 0x10860db70>}})

---- test 4 ----
created context 140232601329568
5.33379434247252
1
defaultdict(<class 'dict'>, {140232600730528: {'Py:ao1': <epics.ca._CacheItem object at 0x10860db70>}})
```
</details>

I believe this will alleviate most of the threading issues people run into using `epics.PV`. 

If we agree this is a good thing, someone should at least double check that I didn't forget to wrap any other important methods on `epics.PV` with `_ensure_context`.

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -9,6 +9,7 @@
 import time
 import ctypes
 import copy
+import functools
 from math import log10
 
 from . import ca
@@ -21,6 +22,44 @@ except ImportError:
     from argparse import Namespace
 
 _PVcache_ = {}
+
+
+def _ensure_context(func):
+    '''
+    Wrapper that ensures a method is called in the correct CA context
+
+    Assumes the instance has a `context` attribute
+
+    Raises
+    ------
+    RuntimeError
+        If the expected context (self.context) is unset (None), or the current
+        thread cannot get a valid context.  Both conditions would normally
+        result in a segmentation fault if left unchecked.
+    '''
+    @functools.wraps(func)
+    def wrapped(self, *args, **kwargs):
+        initial_context = ca.current_context()
+        expected_context = self.context
+        if expected_context is None:
+            raise RuntimeError('Expected CA context is unset')
+        elif expected_context == initial_context:
+            return func(self, *args, **kwargs)
+
+        # If not using the expected context, switch to it here:
+        if initial_context is not None:
+            ca.detach_context()
+        ca.attach_context(expected_context)
+        try:
+            return func(self, *args, **kwargs)
+        finally:
+            # Then revert back to the initial calling context
+            if initial_context is not None:
+                ca.detach_context()
+                ca.attach_context(initial_context)
+
+    return wrapped
+
 
 def get_pv(pvname, form='time',  connect=False,
            context=None, timeout=5.0, **kws):
@@ -163,6 +202,7 @@ class PV(object):
         self._args['chid'] = self.chid = chid
         self.__on_connect(pvname=pvname, chid=chid, conn=conn, **kws)
 
+    @_ensure_context
     def force_read_access_rights(self):
         """force a read of access rights, not relying
         on last event callback.
@@ -253,12 +293,11 @@ class PV(object):
         self.connected = conn
         return
 
+    @_ensure_context
     def wait_for_connection(self, timeout=None):
         """wait for a connection that started with connect() to finish"""
 
         # make sure we're in the CA context used to create this PV
-        if self.context != ca.current_context():
-            ca.attach_context(self.context)
         if not self.connected:
             start_time = time.time()
             if not self._conn_started:
@@ -271,6 +310,7 @@ class PV(object):
                     ca.poll()
         return self.connected
 
+    @_ensure_context
     def connect(self, timeout=None):
         "check that a PV is connected, forcing a connection if needed"
         if not self.connected:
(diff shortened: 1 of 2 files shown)


newville:
@klauer Great idea!   

If I read the intention here, any method that might be called by user code should be wrapped, as the user may have cleverly or inadvertently changed context.  

At a first glance, I wonder if `_on_connect()` should be wrapped too.  This mostly gets called as a callback (which should have a predictable context), but sometimes (say, `force_reconnect()`) this gets called from user code.

klauer:
@newville Thanks! Good point. I wrapped some additional functions based on your comment. Am I still missing any?

newville:
@klauer ... and how about this one... is this ready to go?  

Personally, I think this is a really great idea but I think I am not that expert in threading to see what the unintended consequences might be.   As with #150, I think this is a clear improvement but am unsure if we have a full set of tests for what can go wrong.   So maybe we should merge and live with it as a release candidate.   Does that seem OK?   Does anyone else have any thoughts, concerns, objections?

klauer:
(Rebased/force pushed to resolve merge conflict from the other PRs)

The new decorator will make `epics.PV` instances work in cases where they would fail miserably before. It should effectively be a no-operation for single-threaded operation (or multi-threaded operation with the same context). Keeping the above in mind, I think a release candidate based on this would be fine.

We can definitely revisit this down the line with further testing.

epics/pv.py
```diff
@@ -253,12 +296,11 @@ def __on_connect(self, pvname=None, chid=None, conn=True):
         self.connected = conn
         return
 
+    @_ensure_context
     def wait_for_connection(self, timeout=None):
         """wait for a connection that started with connect() to finish"""
 
         # make sure we're in the CA context used to create this PV
```
klauer: Remove stale comment

newville: yep, I agree (and sorry for delay.  and thanks for closing all those issues. and thanks in general!)

klauer:
No rush, @newville!

I shouldn't neglect to mention: the alternative to this PR could be `_ensure_context` raising an exception if the thread isn't using the specific context that the PV was created in.

In the interest of making it easier for users, though, I think the implementation in this PR (switch context if necessary) is likely superior.

newville:
@klauer  Yes, I agree that this `_ensure_context` wrapper is preferred to raising an exception!

newville:
@klauer this looks good to me.... OK to merge?

klauer:
I think so, @newville.

newville:
@klauer OK, great, I'll merge this.   

Should we then tag that as 3.40.rc2 or should we see if we can fix #145  (we have a proposed approach to a solution, just not code)?

klauer:
@newville hmm, I think we could roll that fix into the next RC. I'll look into it a bit more tomorrow.

PR #152, opened 2019-04-16
MNT: remove undefined raise_on_unconnected

klauer:
(As discussed previously)


newville:
@klauer yep, let's just merge this

Issue #68, opened 2016-09-13
PV can not be reconnected after disconnect


protofabrik:
We've stumbled across a rather inconvenient bug, that is after we disconnect from the PV the PV can not be reconnected anymore. Neither PV.connect nor PV.force_connect work correctly. The first one never connects, the second one successfully connects to the channel but fails with key error when re-adding the callback. See attached files for stand-alone demonstration of the bug... 

``` python
"""
Test-case that demonstrates that after disconnect and channel can not be reconnected using connect method

To try it out, simply start example CAS server bundled with EPICS base (excas binary) before starting the test

The output for me is: 

Connecting
UPD 6.63983507548
SUM 6.63983507548
UPD 76.9632902148
Disconnecting
Reconnecting
Traceback (most recent call last):
  File "test.py", line 33, in <module>
    assert(pv_test.wait_for_connection(5.0) == True)
AssertionError

"""

import epics
import time

pv_test = epics.PV("alan",auto_monitor=True)


def callback(value=None,**kwargs):
    print("UPD",value.sum())



pv_test.add_callback(callback)

print("Connecting")
assert(pv_test.wait_for_connection(5.0) == True)

print("SUM",pv_test.get().sum())

time.sleep(3)

print("Disconnecting")
pv_test.disconnect()
print("Reconnecting")

# Always fails :( 
pv_test.connect()
assert(pv_test.wait_for_connection(5.0) == True)
pv_test.add_callback(callback)


time.sleep(5)
```

``` python
"""
Test-case that demonstrates that after disconnect a channel can be
reconnected using undocumented force_connect, however the callback can not be added again (after force_connect no original callbacks are invoked)

To try it out, simply start example CAS server bundled with EPICS base (excas binary) before starting the test

The output for me is: 

Connecting
UPD 6.92243429227
SUM 6.92243429227
UPD -6.11459633615
Disconnecting
Reconnecting
Traceback (most recent call last):
  File "test2.py", line 49, in <module>
UPD -6.11459633615
    pv_test.add_callback(callback)
  File "/usr/lib/python3.5/site-packages/epics/pv.py", line 502, in add_callback
    self.get_ctrlvars()
  File "/usr/lib/python3.5/site-packages/epics/pv.py", line 425, in get_ctrlvars
    kwds = ca.get_ctrlvars(self.chid, timeout=timeout, warn=warn)
  File "/usr/lib/python3.5/site-packages/epics/ca.py", line 419, in wrapper
[...]

PR #153, opened 2019-04-16
FIX: timeout during get_timevars

klauer:
In the case of timeout, these functions return `None`. Comparing against `None` is necessary prior to attempting to update internal PV state dictionaries.

Also adds a test based on #68. It *appears* to no longer be an issue, but I'm not 100% certain as of yet. More tests should be done.


newville:
@klauer this looks good to me.  Any objections to merging this?

klauer:
I think this PR shouldn't be controversial. Up to you whether you think it warrants closing #68, though.

newville:
@klauer  OK, I'll merge this then.

PR #154, opened 2019-04-25
FIX: auto monitor on reconnect

### LINKED ISSUE #145: No monitor for large PVs after clear_auto_monitor()

klauer:
With this PR, both `PV.clear_auto_monitor` and `PV.reconnect` will not reset the boolean `auto_monitor` - which determines whether or not to set up a monitor automatically on PV connection. 

This means that the one-time logic that set `auto_monitor` upon the initial connection will be re-used for subsequent reconnections (except if the user modifies the attribute directly).

Closes #145

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -323,6 +323,7 @@ class PV(object):
     @_ensure_context
     def clear_auto_monitor(self):
         """turn off auto-monitoring: must reconnect to re-enable monitoring"""
+        self.auto_monitor = False
         if self._monref is not None:
             evid = self._monref[2]
             ca.clear_subscription(evid)
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -233,7 +233,6 @@ class PV(object):
         # occassionally chid is still None (ie if a second PV is created
         # while __on_connect is still pending for the first one.)
         # Just return here, and connection will happen later
-        t0 = time.time()
         if self.chid is None and chid is None:
             ca.poll(5.e-4)
             return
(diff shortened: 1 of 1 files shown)


klauer:
Maybe `clear_auto_monitor` should optionally clear `auto_monitor`? I'm not sure what's the desired outcome there.

newville:
@klauer 

Thanks!  I think that is taking the simpler solution from @dchabot  and @hinxx.  That's fine.  Maybe it would useful to add `auto_monitor` keyword to  `reconnect()`?     I don't have a strong opinion on this.

> Maybe clear_auto_monitor should optionally clear auto_monitor?

Yeah, just from a "what does a method named XXX do", it does seem like that should clear `auto_monitor`, explicitly setting it to `False` as well as clearing actual subscriptions.   Do you agree?    
Again, I don't have a strong opinion.  This is a dramatic improvement as it is.

klauer:
OK, I think you're right, @newville, regarding the `clear_auto_monitor`. I'll change it.

I also have an alternative in mind - I'll open another PR and see what everyone's preference is.

PR #155, opened 2019-04-25
REF: allow changing of PV.auto_monitor at any time

klauer:
A more complicated but more flexible alternative to #154:

* Allows `auto_monitor` to be changed before connection or while connected.
* If set prior to connection, the behavior will be as it is in the current master (checking the count if `None`, etc.)
* If set after connection, the `auto_monitor` is re-evaluated as follows:
   * If newly disabled via `auto_monitor = False` and a monitor exists, then the subscription will be cleared
   * If enabled via `auto_monitor = True` and a monitor does not exist, a monitor will be created
   * If maybe-enabled via `auto_monitor = None` and a monitor does not exist, a monitor will be created
   * If a mask is reconfigured via `auto_monitor = DBE_LOG|DBE_ALARM`, the current monitor will be replaced (or created)

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -148,7 +148,7 @@ class _CacheItem:
     get_results : dict
         Keyed on the requested field type -> requested value
     callbacks : list
-        One or more user functions to be called on change of value
+        One or more user functions to be called on change of connection status
     access_event_callbacks : list
         One or more user functions to be called on change of access rights
     '''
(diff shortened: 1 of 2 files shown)


newville:
@klauer  This looks great -- I think that will be much clearer and more flexible.  

Are we ready to merge this?   Does anyone else have concerns or suggestions?

klauer:
Great 👍. I don't think there _should_ be anything unexpected here. Things have been mostly quiet in the past few PRs, I'm hoping that means others are on board with the changes.

One thing I noticed while tweaking `PV.disconnect` for this was that we never actually call `ca.clear_channel(chid)`. I left a TODO, and testing of that could be done in another PR.

newville:
@klauer  OK, let's merge!

PR #158, opened 2019-05-08
FIX: have get_pv respect callbacks

### LINKED ISSUE #157: Cached PVs do not add callbacks upon further calls to get_pv

klauer:
Closes #157

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -105,7 +105,6 @@ def get_pv(pvname, form='time', connect=False, context=None, timeout=5.0,
     if form not in ('native', 'time', 'ctrl'):
         form = 'native'
 
-    thispv = None
     if context is not None:
         warnings.warn(
             'The `context` kwarg for epics.get_pv() is deprecated. New PVs '
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -10,6 +10,7 @@ import time
 import ctypes
 import copy
 import functools
+import warnings
 from math import log10
 
 from . import ca
@@ -105,15 +106,23 @@ def get_pv(pvname, form='time', connect=False, context=None, timeout=5.0,
         form = 'native'
 
     thispv = None
-    if context is None:
-        context = ca.initial_context
-        if context is None:
-            context = ca.current_context()
+    if context is not None:
+        warnings.warn(
+            'The `context` kwarg for epics.get_pv() is deprecated. New PVs '
+            'will _not_ be created in the requested context.'
+        )
+    else:
+        if ca.current_context() is None:
+            ca.use_initial_context()
+        context = ca.current_context()
 
     pvid = (pvname, form, context)
     thispv = _PVcache_.get(pvid, None)
 
     if thispv is None:
+        if context != ca.current_context():
+            raise RuntimeError('PV is not in cache for user-requested context')
+
         thispv = default_pv_class(
             pvname, form=form, callback=callback,
             connection_callback=connection_callback,
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -63,7 +63,7 @@ def _ensure_context(func):
 
 def get_pv(pvname, form='time', connect=False, context=None, timeout=5.0,
            connection_callback=None, access_callback=None, callback=None,
-           **kwargs):
+           verbose=False, count=None, auto_monitor=None):
     """
     Get a PV from PV cache or create one if needed.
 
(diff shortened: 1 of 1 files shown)


newville:
@klauer I think that's a great idea.  

Does anyone have any concerns about this?

klauer:
@newville it just occurred to me that `get_pv(pvname, context=...)` actually has had no effect on the context that a new PV is created in.

Do you think we should follow through and support that argument or remove it? I don't have a strong opinion - either way you decide.

newville:
@klauer 
> @newville it just occurred to me that get_pv(pvname, context=...) actually has had no effect on the context that a new PV is created in.

> Do you think we should follow through and support that argument or remove it? I don't have a strong opinion - either way you decide.

Good point: a PV is always created in the current context, so `get_pv(...., context=SomeContext)` won't create a PV in a different context -- but it will find an existing one only if the context matches.

We could just enforce that PVs are always created in the initial context.  Then we wouldn't even have to include context in the hash key.  But, that's probably too drastic.  I think having PVs created and looked up for the *current* context is acceptable.  

I guess that means that `get_pv()` should probably always create/return PVs for the current context.   So, it should probably *not* have a `context` function argument -- maybe raising a deprecation warning if passed in as `not None`.

FWIW, I think the code to determine the context should probably match the code in `PV.__init__()`:

``` python
        if ca.current_context() is None:
            ca.use_initial_context()
        context = ca.current_context()
```

does that seem sensible to you?

klauer:
Yes, @newville, your logic is sound there. I deprecated the `context` kwarg in 1ca46af, warning if it is specified. If the PV is not cached and the requested context differs from the current context, `get_pv` will raise an exception stating that it won't create a PV in the wrong context. 

I think that may partly stop users from shooting themselves in the foot... If this passes the tests, I think it should be ready. And maybe time for another RC?

newville:
@klauer  Great, thanks very much!  Yes, I agree this will be a great help.   It looks like tests are passing, so I will merge this and then tag as 3.4.0rc2.   For releasing 3.4.0, I would aim for next week sometime unless there were objections. 

FWIW, there were recent discussions with @brunoseivam  and on Tech-Talk (see https://epics.anl.gov/tech-talk/2019/msg00778.php) about challenges using pyepics with pyDevSup and a python interpreter embedded in an IOC.  The upshot of that discussion might inspire some code changes in how libca is loaded, but it might also demand a switch to epics base 3.15.7 which has not been released yet, or the next version of base 7, which might be even scarier ;).  

Anyway, I'm going to guess that we don't have a complete solution for "pyepics embedded in an IOC" readily in hand and will release 3.4.0 without that fully fixed.  That said, if there are simple fixes (perhaps try to test for "am I embedded in a process that has already loaded libca and, if so, do NOT register the automatic cleanup of CA on ext") that could help, we could try to add them for 3.4.0 (and still get it out by the end of the month!).   Any thoughts or suggestions on how urgent this is?

klauer:
Hmm, as far as pyDevSup + pyepics, it seems that the end result is annoying but not necessarily disastrous - since it happens during cleanup at exit and not randomly during normal operation.
I agree in that if we could add some simple checks/workarounds that would be just fine. 

I haven't actually worked with pyDevSup, though. Maybe someone who works with it regularly would be willing to test/contribute those fixes?

brunoseivam:
w.r.t. pyDevSup, I think the recent fixes in pyDevSup ([this](https://github.com/mdavidsaver/pyDevSup/commit/9237bb4d955152f3eb62162ccf1f9390b30168d2) and [this](https://github.com/mdavidsaver/pyDevSup/commit/c4a1c7deeacbe25ea1edafdda9b770aa3faeaeee)) and in [base](https://code.launchpad.net/~info-martin-konrad/epics-base/+git/epics-base/+merge/366996) are enough to prevent the issue discussed in the tech-talk thread. I have been testing our IOCs here at FRIB and they seem to exit happily!

klauer:
@brunoseivam great! Thanks for the update. 

I guess then we just need to update the pyepics shared libraries when 3.15 gets its next release, @newville?

newville:
@brunoseivam @klauer great to hear!  We won't worry about it for 3.4.0.  

That said, I'm both frightened and excited by the idea of using pyepics with pyDevSup.  It could allow python programs embedded in an IOC to replace seq code.

newville:
@klauer yes, I think we should still plan on including shared libs from 3.15.7 when it is released, so probably for pyepics 3.4.1.    For now, 3.4.0 seems like a big enough change!

PR #160, opened 2019-05-14
BLD: if using NOLIBCA do not install epics.clibs

tacaswell:


newville:
@tacaswell is there a problem having a mostly empty `epics.clibs` directory?  If so, what is that problem?

I am completely at a loss for why anyone would want the NOLIBCA option.  To me, it appears to be an obvious misfeature.  The provided shared libs are what we test with.  We know that they work.  We know they avoid issues that can occur when using "standard epics build" versions of libca/libCom. They are contained within the Python site-packages folder and will not be found by other applications looking for libca.   

If one wants to use a different version of libca, one can set PYEPICS_LIBCA.   When used in a "regular" python environment in one of the tested environments, setting PYEPICS_LIBCA should never be necessary.   It may be necessary if using a very old version of Linux (which should be taken as a hint to not use that version of Linux) or MacOS so old that it is not supported by any vendor (a similar hint).  It probably *is* necessary to set PYEPICS_LIBCA when using pyepics embedded in an IOC -- something we are just starting to consider how to support well.

Unlike setting PYEPICS_LIBCA, installing with NOLIBCA is saying that you *never* want the option to use the known-to-work, tested shared libraries and that you will *always* use an untested version that must be specified by some other environment setting.   I cannot imagine anyone making that choice.

I am reluctant to give the impression that we in any way encourage installing with NOLIBCA.

tacaswell:
> I am completely at a loss for why anyone would want the NOLIBCA option.

We have been over this several time in several places and it is unlikely that _this_ time we will reach any sort of consensus. 

> I cannot imagine anyone making that choice.

Given this context and your previous comments I am not sure it is worth engaging on the merits. 

I would turn the question around: if `epics.clibs` is always empty of python code, is never directly imported, is there any harm in not including it if it does not contain any binaries?

newville:
>> I am completely at a loss for why anyone would want the NOLIBCA option.

> We have been over this several time in several places and it is unlikely that this time we will reach any sort of consensus.

Sorry but I don't recall any arguments other than "trust us, we want it".   Maybe I'm just not remembering all of the discussion (Just to be clear: this is not my full-time job).  I believe that around the time this option was added the behavior of where the shared libraries got installed was different with `pip install` and `python setup.py install` and there *was* an issue of installing to "the wrong place", but that got solved.  Perhaps there was more to it than that. Or perhaps the current behavior means that whatever problem the NOLIBCA option was meant to address is no longer relevant.  Like I said, I don't know why it was needed in the first place, and I still don't know.

> Given this context and your previous comments I am not sure it is worth engaging on the merits.

That doesn't enlighten me much.  Instead of telling me why this option is needed you are telling me that you are not going to tell me. Hm.

> I would turn the question around: if epics.clibs is always empty of python code, is never directly imported, is there any harm in not including it if it does not contain any binaries?

Shrug.  Maybe "if it ain't broke, don't fix it"?  But also, why are you turning the question around instead of just answering it?  Is it too much to bother answering a question about what problem a PR solves?

You made a PR that has no description.  This is a small group of people working on a fairly slow-moving and mature project.  We are not in a mode of rapid development.  We are as worried about unintentionally breaking things as we are about adding features or really even fixing things.  It is reasonable to expect a PR to describe what problem it solves.  It is reasonable to expect some discussion about any PR.  Certainly, *refusing* to give an explanation and instead replying with "well, why is it *not* needed?" as you have done here does not strike me as a good way to make a convincing case.  Maybe I'm not understanding your goal.

So, maybe try again: Is an almost empty clibs folder from installing with NOLIBCA causing a problem?   If so, what is that problem?  Why do we even need a NOLIBCA option?  Those are all honest questions.   I don't know the answer to any of them.

newville:
PRs need a description and to come from people willing to have a meaningful discussion.

PR #161, opened 2019-05-20
FIX: clear channel should remove cache item

### LINKED ISSUE #163: Connection callback does not  work as expected
Hello,
It looks like the connection callback can only get connected event, the pyepics version I'm working with is `3.4.0rc1+26.g5585324`, it used to be working for `3.3.0`. Here is the code that can show this issue:

```python
from epics import PV

def conn_cb(pvname, conn, **kws): 
    print("{} is connected? {}".format(pvname, conn))

pv = PV('test_pv', connection_callback=conn_cb) 
```

And the simple IOC for testing (start with `softIoc -d test.db`):
```
# test.db
record(ao, "test_pv") {
    field(DTYP, "Soft Channel")
    field(OUT, "0")
    field(VAL, "0")
    field(PINI, "YES")
}
```
When the `pv` is created, it outputs `test_pv is connected? True`, but when the ioc is killed, no disconnection event was captured.

klauer:
The overall goal of this PR is to fix `clear_channel` not clearing the cache appropriately. This meant that subsequent calls to `create_channel` after `clear_channel` could return a stale `chid` and cause a segfault. 

Fixing the above uncovered an atexit cleanup bug (with `finalize_libca`) where channels could potentially get cleared twice, causing a segfault.

I believe this PR now addresses both of the above. Additional changes and details are as follows:

* Also adds a simple repr for further debugging of cache-related things.
* Removed the `_get_or_create_*` methods after thinking more about how `create_channel` operates 
    - This simplifies things considerably, and almost everything goes through `get_cache` now.
* Add a `channel id` cache (`epics.ca._chid_cache`).
    - `chid` is effectively a pointer, so we're guaranteed it'll be unique
    - Due to the inner-workings of `caget` and such, we still need the `context + pvname` cache (that is, the user calls `caget` with a pvname, and we need to handle the channel id mapping behind the scenes)
    - I should not have doubted @dchabot previously - it turns out that access rights callbacks are _not always_ called from a thread with a CA context attached. I don't know why this is, but this cache works around it.
   - I believe that there were multiple cache items being created due to the access rights event popping up in a different (i.e., null) context, which created a different cache item. That meant we called `clear_channel` twice at exit.
* Tests were updated, where possible, to use `get_pv` instead of creating their own `PV` instance.
* Last but not least: closes #163 - the new `chid` cache made solving this rather easy. The issue is that disconnection callbacks also do not happen in a CA context, and we did not have test coverage of that.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -78,6 +78,7 @@ DEFAULT_CONNECTION_TIMEOUT = 2.0
 ## Cache of existing channel IDs:
 # Keyed on context, then on pv name (e.g., _cache[ctx][pvname])
 _cache = collections.defaultdict(dict)
+_chid_cache = {}
 
 # Puts with completion in progress:
 _put_completes = []
@@ -154,9 +155,10 @@ class _CacheItem:
     '''
 
     def __init__(self, chid, pvname, callbacks=None, ts=0):
+        self._chid = None
+        self.context = current_context()
         self.lock = threading.RLock()
         self.conn = False
-        self.chid = chid
         self.pvname = pvname
         self.ts = ts
         self.failures = 0
@@ -168,16 +170,29 @@ class _CacheItem:
 
         self.callbacks = callbacks
         self.access_event_callback = []
+        self.chid = chid
+
+    @property
+    def chid(self):
+        return self._chid
+
+    @chid.setter
+    def chid(self, chid):
+        if not isinstance(chid, dbr.chid_t):
+            chid = dbr.chid_t(chid)
+
+        self._chid = chid
 
     def __repr__(self):
         return (
-            '<{} {!r} {} failures={} callbacks={} access_callbacks={}>'
+            '<{} {!r} {} failures={} callbacks={} access_callbacks={} chid={}>'
             ''.format(self.__class__.__name__,
                       repr(self.pvname),
                       'connected' if self.conn else 'disconnected',
                       self.failures,
                       len(self.callbacks),
-                      len(self.access_event_callback)
+                      len(self.access_event_callback),
+                      self.chid_int,
                       )
         )
 
@@ -205,7 +220,7 @@ class _CacheItem:
             if callable(callback):
                 callback(ra, wa)
 
-    def run_connection_callbacks(self, conn, timestamp):
+    def run_connection_callbacks(self, chid, conn, timestamp):
         '''
         Run all connection callbacks
 
@@ -216,9 +231,11 @@ class _CacheItem:
         timestamp : float
             The event timestamp
         '''
-        self.conn = conn
-        self.ts = timestamp
-        self.failures = 0
+        # Lock here, as create_channel may be setting the chid
+        with self.lock:
+            self.conn = conn
+            self.ts = timestamp
+            self.failures = 0
 
         chid_int = self.chid_int
         for callback in list(self.callbacks):
@@ -398,12 +415,16 @@ def finalize_libca(maxtime=10.0):
         start_time = time.time()
         flush_io()
         poll()
-        for ctxid, ctx in _cache.items():
-            for pvname, info in ctx.items():
-                if info.chid is not None:
-                    libca.ca_clear_channel(info.chid)
-            ctx.clear()
+        for chid, entry in list(_chid_cache.items()):
+            try:
+                print('clearing chid', chid, name(chid))
+                clear_channel(chid)
+            except ChannelAccessException:
+                pass
+
+        _chid_cache.clear()
         _cache.clear()
+
         flush_count = 0
         while (flush_count < 5 and
                time.time()-start_time < maxtime):
@@ -412,7 +433,8 @@ def finalize_libca(maxtime=10.0):
             flush_count += 1
         context_destroy()
         libca = None
-    except Exception:
+    except:
+        print('catch-all for exceptions', sys.exc_info())
         pass
     time.sleep(0.01)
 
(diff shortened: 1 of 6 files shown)


klauer:
Hmm, well, I thought this would be a simple PR, but it's passing on 2.7, failing on 3.5/3.6. Will look further into it.

newville:
@klauer Thanks for looking into this -- it looks tricky and painful! 

Out of curiosity, do you have a simple method using pyepics that always generates a segfault on exit?  
I do see these sometimes (maybe even more than "occasionally"), but I'm never convinced they're from pyepics and not other complicated code (say, GUI or database connections).

newville:
@klauer I'd like to release 3.4.0 very soon (ideally today, but perhaps wait until Tuesday).  Is this fixable or should it wait for 3.4.1?

klauer:
Sorry, @newville, I've been out on vacation for the past week. The tests (despite passing) from 5e0e2a5 will repeatably segfault the interpreter at exit. I have a hunch it's due to instantiating with `PV()` instead of using `get_pv()`, but we'll see...

newville:
@klauer OK.  I guess I'm wondering if this should hold up releasing 3.4.0?  I definitely find "segfault on exit" irritating, but also hard to track down and fix.  So, do you think this is urgent or should we release 3.4.0 without this fix.  To be clear, 3.4.0rc2 is passing tests but I can totally believe we're not testing every case that could cause segfault on exit.

klauer:
@newville I think it may be an important part of the next release, and I'd like to fix it. I'm unfortunately running out of spare time to dedicate to hunting down the issue though... Maybe let's wait a few days and I'll see if I can find it. 

Regardless, the previous release (i.e., before my recent cache-related fixes) did not correctly clear the cache on `clear_channel` - so I think even without this PR there would still be significant benefits to the end-user.

newville:
@klauer OK, waiting a few days or a week is fine with me.    FWIW, I'm running 3.4.0rc2 without problems but the summer run at the APS starts tomorrow and I will be doing quite a bit of commissioning, so I doubt I' can help much over the next week. 

Are you able to get predictable and reliable "segfault on exit"?

klauer:
Great to hear you're running it without issue!

I can't reproduce the crashes on my work machine (which is OSX), but they are very reproducible with the test suite on Linux (as you'll see on travis).

newville:
@klauer Sorry, but maybe I'm not understanding something. Aren't the errors only *with* this PR.  The current master branch (558532440ad812e4fd174023a99f44777a97f335) passed all tests on Travis, without segfaults, right? (https://travis-ci.org/pyepics/pyepics/builds/536914574?utm_source=github_status&utm_medium=notification).

And it looks to me like the errors are coming when checking the size of waveform PVs.  

It seems that changing from `PV()` to `get_pv()` in the "waveform with specifically set length" tests might not be working.  That is, it looks like `get_pv()` on an existing waveform PV will use the `count` of the cached PV, not the count specified.   

It's not clear to me whether `count` on a PV should be reset by a `get_pv()`, but maybe.....
Alternatively, `count` could be part of the cache key.   Or maybe the test needs to be changed back to 
use `PV()` instead of `get_pv()`.

klauer:
Yes, @newville - it's within the context of this PR that I was referring to above.

After giving it a good bit more thought and refactoring, I think this is in a better spot. I'd like to fix the connection callback stuff on top of this, if we can agree on the contents. Of course, this is much more than I had initially planned in my 1-line commit!

See the updated PR message (in a few minutes) for more information.

newville:
@klauer Hey Ken,  Thanks for all this.  Maybe we should have a recap of what all changed from 3.3.3 and 3.4.0 and what problems were addressed?  It seems like there was a fair amount of fixes-of-fixes (perhaps especially of the connection logic?), so there might be some confusion about what is "signal" and what is "noise".

Would you be willing to write up a short description what has actually changed significantly since 3.3.3 (or even 3.3.0, if some of this was to repair earlier incomplete repairs)?

newville:
@klauer But also, do you now think this is ready for merging (like, does it fix #163 -- it looks like it should, but I didn't have time today to look at this in detail)?

archman:
> @klauer But also, do you now think this is ready for merging (like, does it fix #163 -- it looks like it should, but I didn't have time today to look at this in detail)?

I've tested my app with this PR, #163 is fixed, thanks!

klauer:
Yes, @newville, I added a test for the disconnection callback. @archman has confirmed that it works as well.

I'll see what I can do about a detailed changelog, but the general gist of it is something along the lines of: 
* improved caching mechanism with better thread-safety (but not perfect) for access of PVs and related caget-style utilities
* `get_pv` respects callback parameters for already-existing PV instances (and best-effort attempts to respect other parameters)
* Allow`PV.auto_monitor` to be changed at any point

Some of these are features, but most of it should just be under-the-hood fixes that users shouldn't have to worry about.

klauer:
A couple things were brought to my attention which I'm trying to address with some follow-up commits. I know there are people waiting on the results of this PR - sorry.

I need a bit more time to fully think this through and make it a bit more robust.

newville:
@klauer Sorry for delay, I was tied up with other stuff all day.   It sounds like the consensus is that this should be merged and labelled as 3.4.0 rc3.   Great!  I'll do that now.  If there are any objections or other things for releasing 3.4.0, please let me know.  Hopefully we can release that soon.

For release notes/changelog, the bullet list you have looks like a great start, and I don't think it needs to be a lot more extensive than that. Like you say, the changes are mostly under the hood or at least tweaks to the API (like changing `PV.auto_monitor`) that was not clearly allowed or disallowed before.  So it does seem mostly like "only" bug fixes even if the bugs are actually the most pernicious and hard-to-fix species and the fixes definitively real work.

PR #162, opened 2019-05-29
BLD: if using NOLIBCA do not install epics.clibs

tacaswell:
Replaces PR #160 which was erroneously closed.

Opinions differ if pyepics should be shipping pre-built binaries directly in the package or not.  If we accept the position that there exist reasonable people do not want the binaries (which is what I thought the consensus in https://github.com/pyepics/pyepics/pull/83/ was), then pyepics should not be installing vestigial empty sub-modules in the case where they are not included.

I would like this merged for 3.4.0

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -232,11 +232,15 @@ def _find_lib(inp_lib_name):
             os.path.isfile(dllpath)):
         return dllpath
 
-    # Test 2: look in installed python location for dll
-    dllpath = resource_filename('epics.clibs', clib_search_path(inp_lib_name))
+    try:
+        # Test 2: look in installed python location for dll
+        dllpath = resource_filename('epics.clibs',
+                                    clib_search_path(inp_lib_name))
 
-    if (os.path.exists(dllpath) and os.path.isfile(dllpath)):
-        return dllpath
+        if (os.path.exists(dllpath) and os.path.isfile(dllpath)):
+            return dllpath
+    except ModuleNotFoundError:
+        pass
 
     # Test 3: look through Python path and PATH env var for dll
     path_sep = ':'
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,10 @@ nolibca = os.environ.get('NOLIBCA', None)
 if nolibca is None:
     pkg_data = {'epics.clibs': ['darwin64/*', 'linux64/*', 'linux32/*',
                                 'linuxarm/*', 'win32/*', 'win64/*']}
+    extra_pkgs = ['epics.clibs']
 else:
     pkg_data = dict()
+    extra_pkgs = []
 
 PY_MAJOR, PY_MINOR = sys.version_info[:2]
 if PY_MAJOR == 2 and PY_MINOR < 6:
@@ -72,7 +74,7 @@ setup(name = 'pyepics',
                       'Programming Language :: Python',
                       'Topic :: Scientific/Engineering'],
       packages = ['epics','epics.wx','epics.devices', 'epics.compat',
-                  'epics.autosave', 'epics.clibs'],
+                  'epics.autosave'] + extra_pkgs,
       package_data = pkg_data,
      )
 
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -1,6 +1,6 @@
 import pytest
 import os
-import sys 
+import sys
 import re
 
 import epics
@@ -12,14 +12,14 @@ pytestmark = pytest.mark.skipif(os.getenv('TRAVIS') is None,
 
 def test_libca_init():
     assert epics.ca.initialize_libca()
-    
+
 # Skip if Travis defines NOLIBCA
 @pytest.mark.skipif(os.getenv('NOLIBCA') is not None,
                     reason='NOLIBCA is defined')
 def test_install_encapsulation():
     libca = epics.ca.find_libca()
     system = sys.platform
-    
+
     if 'linux' in system:
         assert re.search('.*/epics/clibs/.*/libca\.so$', libca)
     elif 'darwin' in system:
@@ -33,14 +33,5 @@ def test_install_encapsulation():
 @pytest.mark.skipif(os.getenv('NOLIBCA') is None,
                     reason='NOLIBCA is not defined')
 def test_install_no_libs():
-    libca = epics.ca.find_libca()
-    system = sys.platform
-
-    if 'linux' in system:
-        assert re.search('.*/epics/clibs/.*/libca\.so$', libca) is None
-    elif 'darwin' in system:
-        assert re.search('.*/epics/clibs/.*/libca\.dylib$', libca) is None
-    elif 'win' in system:
-        assert re.search('.*\epics\clibs\.*/ca\\.dll$', libca) is None
-    else:
-        raise 'Are you using BSD? You are hardcore...'
+    with pytest.raises(ImportError):
+        import epics.clibs


newville:
> replaces PR #160 which was erroneously closed.

Hm, I don't think that closure was actually an error.  The PR had no description.  One was asked for and not provided. 

> Opinions differ if pyepics should be shipping pre-built binaries directly in the package or not. 

Well, "opinions differ" on whether the earth is flat ;). The question is what we want to ship.  
We have always provided binary versions of libca for platforms that need them.  This has always included Windows, sometimes included MacOS, and recently came to include Linux.  It is a feature of pyepics that the shared libraries will be included that *should* work and to also provide an easy way to use another version of these libraries if desired.  To be clear, it has also always been an important feature of Epics CA that library versions used by servers and clients do *not* need to match closely, though some features will need versions of the libraries that support those feature. 

> If we accept the position that there exist reasonable people do not want the binaries 
> (which is what I thought the consensus in #83 was), 

By the summer of 2017 we had decided that providing binaries for all systems, including Linux, was necessary.  This mostly due to potential conflicts with libreadline.  To be clear, this conflict is  TOTALLY the fault of epics base in that by default, libCom is linked to libreadline on Linux, and libca linked to that, meaning that libca depends on libreadline though it absolutely does not need it.  Since the Python REPL may also link to libreadline (and, you know, actually *use* it), there is the potential for conflicts -- the version of libreadline used sensibly for Python must match that used non-sensically for libCom.  With multiple version of Python environments (especially say Anaconda, Enthought, etc as opposed to "system-supplied" Python) in wide-use, it was not hard to find situations where pyepics connecting to libca would crash Python.  So, we *must* work around this by building and providing libCom compiled without libreadline.   Of course, if someone wants to use a different version of libca/libCom they always can.

Anyway, in the summer of 2017, we were trying to figure out the best way to include and distribute libca/libCom on all platforms we support.   One problem we faced at that time was that different installation methods (`pip`, `python setup.py install`) would put the binaries in different places depending on the install method, with (IIRC) `pip install`sometimes putting them in `usr/lib` or other places that could potentially overwrite versions installed by other software.  I believe #83 was really a response to that.  I could be mistaken about that - it was a long time ago.

By the time we released 3.3.0 in Oct 2017 we had moved all the binaries inside the package directory, in `epics/clibs` as package data and thereby simplified the search for libca and preventing any collisions with other versions of libca/libCom.  I believe we have made a few tweaks to this since the initial release of 3.3.0.  

In any event, the addition of the NOLIBCA install option came *before* this moving of the install location of the libraries.  I believe the need for NOLIBCA vanished with 3.3.0.  The option still exists, though I do not know why.  Apparently, some people are using it.  I do not know why.   It is entirely possible everyone who thinks they need it actually is wrong and does not understand how pyepics works since 3.3.0.  As I have said before, I have never heard a compelling reason for NOLIBCA. I have asked many times.  I have explained why I think we don't need this many times.  The conversation has become sort of tiresome.   If you or anyone else wants to try to explain why NOLIBCA is needed,  the time to do that is NOW, here, and in writing and here.  Persuade me that this is needed.

> then pyepics should not be installing vestigial empty sub-modules in the case where
>  they are not included.

Is an almost empty clibs folder causing problems?  If so, what are these problems?

> I would like this merged for 3.4.0

Shrug.  I don't know why we need this.  I think removing NOLIBCA may be the better option, though I do not find that urgent.  

Anyway, the changes for 3.4.0 strongly reflect Ken's work on better caching, and we're trying to figure out a few odd test failures and #161 and #163.  We're well past 3.4.0rc2.   This PR is not related to that and definitely does not seem urgent.  So unless I become convinced by the need for this (and quickly), I doubt this will make it into 3.4.0.

tacaswell:
You are describing what is  fundamentally a packaging problem.  

Bundling binary blobs of pre-built libca/libcom in the source tree is one way of solving the packaging problem, `NOLIBCA` makes it easier to solve the packaging problem in other ways.

klauer:
I've tried to stay out of these discussions, but I will say one thing:
I think that even if there's a fundamental disagreement on its utility, including this PR in pyepics will not have an effect on anyone but those who explicitly try to use it. It's also a simple enough change that we should be relatively confident it will not have any side effects.

newville:
@tacaswell    Yep, this is a packaging issue.   That does not actually answer my question.

@dchabot 
>> I have asked many times. I have explained why I think we don't need this many times. The conversation has become sort of tiresome. If you or anyone else wants to try to explain why NOLIBCA is needed, the time to do that is NOW, here, and in writing and here. Persuade me that this is needed.

> Your argument seems to be, "I don't need this mechanism, therefore it should not be available to anyone". This is a non-sequitur, and leads to the circular, frustrating-for-everyone discussions we've had in the past.

Actually, my argument is not that I don't need it so no one needs it.  My argument is that I don't know why anyone ever needs this.  And let me be clear because some people in this conversation seem to believe that I have some devious motivation or something.  I do not.  I really, do not know why anyone needs this.  I've asked a number of times.  You all are saying you want it, and still *not* saying *why*.    Really, what problem is solved by not having default shared libraries?  

You are saying that it is an escape hatch.  Well, from what?  What is the problem to be avoided?  If you're having problems with the packaged libca, let's solve them. 

Installing with NOLIBCA means that the *installer* decides that all user of that install of pyepics to NEVER be able to use the default (and tested) shared libraries.  That doesn't seem like an escape hatch to me, it seems more like a hole that forces the user to specify how to find libca.  

If you are looking for an escape hatch, that is PYEPICS_LIBCA.

> "But, can't you just use PYEPICS_LIBCA?"
> No. That mechanism makes the sourcing of libca a user responsibility.

Well, installing with NOLIBCA also makes it a user responsibility. Any method that relies on environmental variables will be the users responsibility.  Of course, they can be front-loaded at a system-level if desired or auto-inserted into a bash or whatever shell startup script.    But when installed with NOLIBCA, the user must have environmental variables (PATH, LD_LIBRARY_PATH etc) set in a such a way as to find libca.   That can definitely cause problems.  OTOH, I don't know of any problems using the versions in epics/clibs. 

If the goal is to have a custom version of libca for some other packaging system and you want to be certain which version of libca is used, you would be better off overwriting the libca in epics/clibs with your own choice.  Pertinent to this PR, you would not blow away `epics/clibs`, you would *use* it.  That is the most reliable way to be certain which library the user was loading (you might also unset PYEPICS_LIBCA).  All other methods rely on environmental variables that can be different per shell instance.  

Then again, most packaging systems I've ever worked with are all about the pre- and post-install patches.  So if you're writing packaging scripts for some packaging system you can probably install normally (ie, without NOLIBCA) and then alter the installed package anyway you want, including replacing the shared libraries.    
 
Anyway, I agree that NOLIBCA is mostly unobtrusive. Does it actually do anything useful?  I don't know.  

I don't know why it is important to remove a mostly empty `epics/clibs`.  Is that causing problems?

klauer:
Am I understanding the order of library path searching correctly?

  1. `PYEPICS_LIBCA`
  2. pyepics (`epics.ca.*`)
  3. pyepics package path (`epics.ca.__file__`)
  4. pyepics package parent path (`epics.ca.__file__/../`)
  5. Python `lib` path (`os.__file__`)
  6. Parent directory + `DLLs/` (or `lib/`, depending on platform)
  7. Python environment root (`sys.prefix`)
  8. `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, ...

I'm sure I confused at least one of the above - it's a bit obtuse in `_find_lib`. 

I can see how it would be frustrating for `LD_LIBRARY_PATH` to be set in such a way that every `libca`-based application but pyepics would pick it up because it short-circuits.

Maybe if we at least re-ordered the priority of these, perhaps others wouldn't feel the need to push so hard for `NOLIBCA`? I don't feel strongly enough about this to push for it, but I do see where others are coming from here.

newville:
@klauer We could certainly alter the default search path used for libca.   

But, I have to caution you that the notion that pyepics *should* use the same version of libca as other CA applications can definitely cause problems.  This is actually what got us into the situation of deciding that we needed to supply our own versions.  I'm pretty sure I've said this before, but I'll try to explain it again:

By default on Linux, libCom (in base 3, don't know about 7) is linked against libreadline.  There is no way to sugar-coat this: this is BRAIN-DEAD.  If one uses such a version of libCom with a standalone Python linked against a different version of libreadline, they will not be able to use pyepics at all.  We cannot control which version of libreadline Python from anaconda, enthought, whatever use.  We cannot control which version of libreadline the "system libCom" is built with.   We have seen these conflicts.  The way to prevent such a conflict is to build and distribute libca and libCom without libreadline enabled.  Installing with NOLIBCA throws this information.  It also makes the search for libca much harder and less predictable.

It's also simply untrue that it is somehow "better" or "good practice" for all epics applications to use the same version of libca.  It is a very intentional and important feature of CA that different versions of libca will work together.  It is truly and profoundly anti-important that there is a single version of libca.   I understand that this is unusual and not normal "good practice" on modern posix-like systems, but Epics is odd in many ways as a Unix-like project, and this is one of those ways (the odd conventions for cross-compiling and the horrific configuration system are probably more egregious).   

Providing our own version of libca/libCom that cannot be found by other applications WILL NOT cause any problems.  No problems have been reported when using the supplied libraries that would be fixed by using system-versions of libca.  The reverse is not true.  

So, the search path used when PYEPICS_LIBCA and `epics/clibs` both fail to identify a libca could certainly be altered.   I don't really have a preference for how that path is defined - it should include LD_LIBRARY_PATH and PATH, but other than that, it doesn't much matter to me.   This will always represent a last ditch effort to find libca, and if you get to this point in the code, the probability of irreparable problems has already gone way up:  The ship is sinking fast and you're grasping at straws.
It is truly unwise to allow users to install with NOLIBCA.  We should not be helping them do that to themselves.

klauer:
OK, I think we've discussed this enough. Sorry, @tacaswell.

PR #164, opened 2019-05-30
TST: run sg_test on CI

tacaswell:


newville:
@tacaswell  Provide a description, please.

Is anyone actually using synchronous groups?  If so, please see https://epics.anl.gov/tech-talk/2018/msg01666.php and the warning at http://pyepics.github.io/pyepics/ca.html?highlight=sg#synchronous-groups.  I think that they make sense only without preemptive callbacks (FWIW, I always wanted the opposite of "preemptive callbacks" to be "emptive callbacks", but it never caught on ;)).  I believe we do absolutely no testing of running without preemptive callbacks which would require explicit pend_io() and pend_event() all over the place -- not a pretty sight.

It has been a long time since I looked at this or even considered using `sg` groups or non-preemptive callbacks, but even then I am not sure they actually really did what they claimed (or at least the naive and most obvious reading of what they claimed), at least not from Python.

Anyway, the tests added here test only that the calls run, not whether the gets and puts are actually synchronous.   I believe there is no guarantee that they will be.

tacaswell:
This did not add any tests, it fixed existing tests that had bit-rotted due to API changes to `_unpack` made ~7 years ago and are discovered by pytest.

If they are not used / useful can they be ripped out instead?

klauer:
I actually I tried to get it working a few years back and found that nothing worked - not at all.

My opinion is that the signal group stuff should be ripped out of pyepics altogether.  If someone finds a need for it down the line, we can worry about it at that point.

newville:
@tacaswell  @klauer yes, I think this test should be removed.  I'd be OK with removing the code for `sg_*`.  It was included for the sake of completeness, but I think it never worked correctly (even in C) with preemptive callbacks.

In the future, please add a description to any PRs.

klauer:
Would be happy to merge a cleanup PR removing `sg_*`. Adding tests for non-working Python code that's not advised for use even in epics-base isn't worth any of our time, though.

newville:
@klauer thanks, yeah probably we should deprecate all the `sg

PR #165, opened 2019-05-31
FIX: re-check context_cache after sentinel lock

klauer:
Grab the entry again from the cache. Between the time the lock was attempted and acquired, the cache may have changed.

Follow-up to the never-ending #161

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -969,6 +969,9 @@ def create_channel(pvname, connect=False, auto_cb=True, callback=None):
     entry = context_cache.setdefault(pvname, _create_channel_sentinel)
 
     with entry.lock:
+        # Grab the entry again from the cache. Between the time the lock was
+        # attempted and acquired, the cache may have changed.
+        entry = context_cache[pvname]
         is_new_channel = entry is _create_channel_sentinel
         if is_new_channel:
             callbacks = [callback] if callable(callback) else None
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -107,8 +107,6 @@ class _SentinelWithLock:
     def __init__(self):
         self.lock = threading.Lock()
 
-_create_channel_sentinel = _SentinelWithLock()
-
 
 class ChannelAccessException(Exception):
     """Channel Access Exception: General Errors"""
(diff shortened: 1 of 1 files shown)


newville:
@klauer   Oh, wasn't the `_SentinelWithLock` meant to prevent that?   Would it work to do that lock acquisition in a single step,  something like

```
    with context_cache.setdefault(pvname, _create_channel_sentinel).lock:
        entry = context_cache[pvname]
        is_new_channel = entry is _create_channel_sentinel

```
?

It does seem like it would be helpful (and probably make the code simpler) to ensure that while running the create_channel() in one thread that other threads waited to create any other channels.   Would it work for `create_channel()` to start with `lock.acquire()`  and make sure to run `lock.release()` for that singleton sentinel lock?    I guess that is what the `with entry.lock` is meant to do...  But it seems slightly complicated to me that `entry` changes within the `with entry.lock` block.

Anyway, that's just a suggestion, and maybe you've already gone through all the simple-minded approaches.

newville:
@klauer Just to follow up, would something like this work for protecting create_channel() being called simultaneously from 2 threads:

```python

@withCA
def create_channel(pvname, connect=False, auto_cb=True, callback=None):
    context_cache = _cache[current_context()]

    with _create_channel_sentinel.lock:
        entry = context_cache.get(pvname, None)
        if entry is None: # new channel
            chid = dbr.chid_t()
            callbacks = [callback] if callable(callback) else None
            entry = _CacheItem(chid, pvname=pvname, callbacks=callbacks)
            context_cache[pvname] = entry

            ret = libca.ca_create_channel(ctypes.c_char_p(STR2BYTES(pvname)),
                                          _CB_CONNECT, 0, 0, ctypes.byref(chid))
            PySEVCHK('create_channel', ret)
            _chid_cache[chid.value] = entry

        elif (callable(callback) and callback not in entry.callbacks):
            entry.callbacks.append(callback)
            if entry.chid is not None and entry.conn:
                # Run the new connection callback if already connected:
                callback(chid=_chid_to_int(entry.chid), pvname=pvname,
                         conn=entry.conn)
    if connect:
        connect_channel(entry.chid)
    return entry.chid
```

I think that is what is intended, but I am not sure if I am missing something.

klauer:
Overall, you have the right idea @newville, it's just that I'm attempting to make the locks a bit more granular. 

With the latest commit I just pushed, the granularity is at the level of lock per specific pvname in a specific context, instead of one lock for the full context (or for that matter, one lock for pyepics globally).

I like the single-step approach you suggested and included it in the last commit as well.

newville:
@klauer  Thanks, that looks OK to me.  Anyone have any objections to merging this?

klauer:
I think this is good, @newville, ready when you are.

newville:
@klauer OK, and great!!   I'll merge this now, then make a separate PR to address #171

PR #168, opened 2019-06-02
New wx class PVFloatText

vstadnytskyi:
## Description
There are 2 static text fields available: PVText and PVStaticText. However, they are text fields and are not very useful to show numeric data. It shows too many unnecessary decimals. Hence I propose to introduce a new field which specifically designed to show numeric data and can have an extra input parameter 'precision' to specify how many decimals to show.


newville:
@vstadnytskyi    For displaying floating values as text strings with "proper precision", the normal approach is to set the Precision field on the PV.  That will cause the string value (which is really what will be displayed in the StaticText label) to be formatted as desired.

An advantage of this approach is that the Precision field is typically exposed as a PV (say Thing.PREC) and so can be set either at PV creation or dynamically. 

That is to say, PVStaticText *ought* to work (I use it all the time) as long as you have the precision set correctly for the PV.

vstadnytskyi:
I guess I am not using Channel Access in a conventional way. I create my own non-standard PVs such as Thing.threshold (or Thing.mean, Thing.stdev) and the precision on it might be different from the precision on Thing.RBV. 

Also, it is a nuisance if the precision of measurement is high and I don't want to see all decimals, but I want to archive all decimals. Hence, I get in a situation when different clients require different precision of a give PV. Is there a solution with a given infrastructure?

newville:
@vstadnytskyi 
>  Is there a solution with a given infrastructure?

I don't think there is.  Basically, a records precision field exists exactly to be the natural precision to display it at, and can be set independent of value.   The value (for a float PV) will be a double precision floating point number.  The `precision` field is used only to generate the string representation of that value.  You can certainly display at a different precision if you want to, but that is sort of why 'PREC' exists.  So I don't know that it really makes sense to have a `PVStaticText()` class that shows floating point PVs just fine using the records resolution and a different `PVFloatText()` that shows a value at user-defined precision.  For many use-cases, PREC *is* the user-defined precision.

newville:
@vstadnytskyi  Can you separate the fix for PVEnumChoice from the rest of this?

vstadnytskyi:
@newville . This is an old pull request, the one that was unsolicited. I have not created a new one mentioned in #193 to make sure it will be a solicited one.

vstadnytskyi:
this was an unsolicited pull-request I have submitted some time ago. This is no longer relevant.

PR #172, opened 2019-06-05
FIX: replace connection callback poll() with sleep()

### LINKED ISSUE #171: Multiple disconnects result in "CAC: TCP circuit creation failure because "Too many open files"" error and Segmentation fault (core dumped)
I am writing a health monitor, therefore this code must be robust and handle misbehaving ioc's. During testing I'm using pcispy to simulate a sioc that repeatedly crashes. The pcispy test script runs for two minutes serving two PV's then exits. A bash script waits a couple of minutes, then restarts it.

My health monitor code monitors the PV's on the pcispy test ioc test fixture. It uses ca.create_channel(pvname, callback = onConnectionChange) to watch for connection changes. Currently, the callback does nothing but report the disconnect (and I cannot find any documented advise on how to handle disconnects).

Currently, it runs fine for the first 1000ish connect/disconnects. Unfortunately, it then gives me the, "Too many open files" error, and Segfaults. Not surprisingly, my ulimit is 1024. I'm running on Ubuntu 18.04 / python3.

It seems that either there is some documentation that I'm not finding or pyepics is not handling disconnects cleanly.

Thank you, tom.c

klauer:
Closes #171

libca apparently(erroneously) leaks UDP sockets when calling `poll` in a connection callback thread. More details and debugging information in #171.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -252,7 +252,10 @@ class _CacheItem:
         chid_int = self.chid_int
         for callback in list(self.callbacks):
             if callable(callback):
-                poll()
+                # The following sleep is here only to allow other threads the
+                # opportunity to grab the Python GIL. (see pyepics/pyepics#171)
+                time.sleep(0)
+
                 # print( ' ==> connection callback ', callback, conn)
                 callback(pvname=self.pvname, chid=chid_int, conn=self.conn)
 


newville:
@klauer Great - thanks!   OK, is this now ready to be 3.4.0?

klauer:
I don't have anything else planned, at least. It was good timing that we learned about this issue, at least.

PR #175, opened 2019-06-20
Fix problem with listing attributes in python3 in __dir__() method

schlepuetz:
## Description

The __dir__() method of epics.Device() was concatenating dictionary keys and lists. This is fine under python2, but broke in later versions of python3. Convert the dict_keys iterables to lists to fix the problem transparently for py2 and py3. Maybe there is a better solution for this, but the quick fix in this pull request seems to solve the problem for now.

This solution has not yet been tested on all versions!

I have not checked whether similar problems would potentially arrise in other modules.

### DIFF
--- a/epics/device.py
+++ b/epics/device.py
@@ -303,9 +303,9 @@ class Device(object):
 
     def __dir__(self):
         # there's no cleaner method to do this until Python 3.3
-        all_attrs = set(self._aliases.keys() + self._pvs.keys() +
+        all_attrs = set(list(self._aliases.keys()) + list(self._pvs.keys()) +
                         list(self._nonpvs) +
-                        self.__dict__.keys() + dir(Device))
+                        list(self.__dict__.keys()) + dir(Device))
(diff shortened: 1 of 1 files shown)


newville:
@schlepuetz Thanks, and great to hear from you.  And, yes, that explicit and immediate conversion of the keys into a list looks fine to me - the keys won't be huge and we do want them all right now.

Anyone else have any objections to merging this?

newville:
@klauer Do you know why the Travis tests are failing for this?  It looks like it could be an error with setting paths for the docker image or some epics base config so that it cannot find the command-line `caget`.  

I have a hard time believing the errors are because of this PR.

klauer:
Looks like the lack of epics-base tools `caget`, and `softioc` are causing the failures. epics-base        
 is from `lightsource2-tag/linux-64::epics-base-3.15.6-0` - something must have changed significantly there...

newville:
@schlepuetz @klauer Thanks!

PR #179, opened 2019-09-06
fix `pv_name` definition supported by autosave

newville:
This is a fix for Tech-Talk https://epics.anl.gov/tech-talk/2019/msg01425.php, and following the recommendations in that discussion for what a valid PV name should be.  

As it turns out, pyepics does essentially no enforcement or expectation of what a valid PV name is except in the sense of "accepted by `ca_create_channel()`", except in the `autosave` module where files of PV names are parsed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ before_install:
   - conda config --add channels lightsource2-tag
 
   # MAKE THE CONDA RECIPE
-  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION epics-base
+  - conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION epics-base=3.14.12.6
   - source activate $CONDA_ENV
 
 install:
--- a/epics/autosave/save_restore.py
+++ b/epics/autosave/save_restore.py
@@ -21,9 +21,9 @@ Web site is http://pyparsing.wikispaces.com/
 
 """
 
-from pyparsing import Literal, Optional, Word, Combine, Regex, Group, \
-    ZeroOrMore, OneOrMore, LineEnd, LineStart, StringEnd, \
-    alphanums, alphas, nums
+from pyparsing import (Literal, Optional, Word, Combine, Regex, Group,
+                       ZeroOrMore, OneOrMore, LineEnd, LineStart, StringEnd,
+                       alphanums, alphas, nums, printables)
 
 import sys
 import os
@@ -178,12 +178,12 @@ float_number = Combine( integer +
                         Optional( point + Optional(number) )
                         ).setParseAction(lambda t:float(t[0]))
 
-# (originally I had pyparsing pulling out the $(Macro) references from inside names
-# as well, but the framework doesn't work especially well without whitespace delimiters between
-# tokens so we just do simple find/replace in a second pass
-pv_name = Word(alphanums+":._$()")
+# PV names according to app developer guide and tech-talk email thread at:
+# https://epics.anl.gov/tech-talk/2019/msg01429.php
+pv_name = Combine(Word(alphanums+'_-+:[]<>;{}')
+                  + Optional(Combine('.') + Word(printables)))
+pv_value = (float_number | Word(printables))
 
-pv_value = (float_number | Word(alphanums))
 pv_assignment = pv_name + pv_value
 
 comment = Literal("#") + Regex(r".*")


newville:
@klauer the Travis failures look like maybe the docker image isn't getting some PATH set correctly, so that `caget`, `softIoc`, etc are not found.   Any ideas?

klauer:
@newville looks like this is probably at the root of it: https://github.com/NSLS-II/lightsource2-recipes/commit/a54a376d8d7941d76d8881c5675d23e25a649461#diff-950d156360fb3fa6d4e94899472bd2a8

My 2c is that it may be wise to just pin to the previous version and avoid the headaches of further CI debugging.

newville:
@klauer yep that looks like the problem  I guess this was seen with #175 too and I forgot.

So what is the right solution?  Should we be using a different docker image?

Or should we add the equivalent of the removed:

```
# Setup symlinks for utilities	
BINS="caget caput camonitor softIoc caRepeater cainfo"	
cd $PREFIX/bin	
for file in $BINS ; do	
	ln -s ../epics/bin/$EPICS_HOST_ARCH/$file .	
done	
```

to the Travis config?   Could that as simple as

```
    export PATH=/home/travis/mc/bin:/home/travis/epics/bin/linux-x86_64/:$PATH
```

or maybe PREFIX is set to something more complicated..?

klauer:
Fixing the path and LD_LIBRARY_PATH could work.

However, my recommendation would be just to pin to the old-and-working version I noted above to save yourself the headache of debugging this.

newville:
@klauer 

> However, my recommendation would be just to pin to the old-and-working version I noted above to save yourself the headache of debugging this.
  
Sorry for being dense, but I don't know how to do that.

klauer:
No worries, @newville. It should be pretty simple:

On its own: `conda install -c lightsource2-tag epics-base==3.14.12.6`

Or modifying the travis configuration: `conda create -n $CONDA_ENV python=$TRAVIS_PYTHON_VERSION epics-base==3.14.12.6` (note it's just the `==3.14.12.6` portion that needs to be added)

newville:
@klauer Thanks!  changing the travis config to specify epics_base worked!

Does this fix for PV names seem OK to you?

klauer:
Great!

It's tough to confirm the accuracy of the grammar for the pv name, but one thing that jumped out to me was the field name, which you have as `pyparsing.printables`:
```
0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&'()*+,-./:;<=>?@[\\]^_`{\|}~
```

I think that's too lax. Perhaps anything that's valid for the record name is also valid for the field name - with 0 or 1 `.` as the `RECORD.FIELD` delimiter?

newville:
@klauer I believe the Andrew Johnson was saying that the FIELD part really can be any printable character as FIELD essentially supports JSON.   

It is kind of weird that "what is a valid PV name" has such an unclear answer.   And like, no one even brought up the question of unicode.  PVName is probably really `char[40]` with downstream assumptions about having one '.' to separate Record from Field.
 
For { and }, yeah I remember thinking "I didn't know that was allowed" the first time I visited NSLS-II.   I did not rush home and say "ooh, that's a nice idea"!  

But, I also remember many years ago trying to use a tool from another facility that turned out to not support PV names that start with numbers, assuming they were like variable names so had to start with letters.  Not a bad assumption.... except of course that almost all of the PV Names I use start with '13'.  Yeah, it is kind of odd.

klauer:
Agreed, @newville - it is all rather strange.

In the context of autosave, I'm not sure you'd use a [channel filter](https://epics.anl.gov/base/R3-15/5-docs/filters.html) on those PVs, but then again who knows 🤷‍♂. I'm OK with merging.

newville:
@klauer On Channel Filters: yeah, I'm not even sure one can put to a JSON-filter. It does look like autosave does not support them.  But, it probably won't hurt anything to have them here....

ralphlange:
There is a distinction:
Channel Access (as the network protocol) puts very little restrictions on the PV name.
The application behind the server (e.g. an EPICS Database on an IOC) may use or enforce its own restrictions.

The EPICS IOC database currently limits the length of a record name to 63 characters (since long ago) and documents a restriction in the allowed set of characters that it doesn't enforce, as discussed above. With the added dot, field name and JSON filter specifications (that are all part pf the Channel Access PV name), that PV name can be much longer.

newville:
@klauer OK, I'll merge this.  I think the main point was to lift the overly restrictive naming rules.  I'd rather go over what a valid PV name is, or what the Autosave module supports, rather than have this module be the limiting factor.

Issue #187, opened 2019-12-03
Motor and Device `get()` should have `timeout` exposed


newville:
See https://epics.anl.gov/tech-talk/2019/msg01858.php

Basically `Motor.get()` should pass a `timeout` argument to the call to `PV.get()`.

PR #188, opened 2019-12-03
pass  from Device/Motor.get() to PV.get() 

newville:
This addresses #187


newville:
@klauer  Great, thanks for looking that over.  I'll merge.

PR #190, opened 2020-01-09
explicitly unpack access bits, as required for Py3.7.6

### LINKED ISSUE #189: Incompatibility with Python 3.7.6
Python 3.7.6 was released on Dec 11th and is incompatible with pyepics, failing on import. Steps to reproduce:
```
$ git clone git@github.com:pyepics/pyepics.git
$ cd pyepics
$ conda create --name py375 python=3.7.5
$ conda create --name py376 python=3.7.6
$ conda activate py375
$ python
>>> import epics
$ conda activate py376
$ python
>>> import epics
  File "/u1/zlentz/conda/envs/py376/lib/python3.7/ctypes/__init__.py", line 97, in CFUNCTYPE
    return _c_functype_cache[(restype, argtypes, flags)]
KeyError: (None, (<class 'epics.dbr.access_rights_handler_args'>,), 1)

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/afs/slac.stanford.edu/u/lu/zlentz/pyepics/epics/__init__.py", line 29, in <module>
    from . import ca
  File "/afs/slac.stanford.edu/u/lu/zlentz/pyepics/epics/ca.py", line 761, in <module>
    dbr.access_rights_handler_args)
  File "/afs/slac.stanford.edu/u/lu/zlentz/pyepics/epics/dbr.py", line 339, in make_callback
    return ctypes.CFUNCTYPE(None, args)(func)
  File "/u1/zlentz/conda/envs/py376/lib/python3.7/ctypes/__init__.py", line 99, in CFUNCTYPE
    class CFunctionType(_CFuncPtr):
TypeError: item 1 in _argtypes_ passes a struct/union with a bitfield by value, which is unsupported.
```
Using
```
python                    3.7.6                h357f687_2    conda-forge
```

newville:
## Description

This changes the ctypes structure definition for `access_rights_handler_args` to explicitly unpack the read/write access bit fields.

An attempt to fix #189

### DIFF
--- a/epics/dbr.py
+++ b/epics/dbr.py
@@ -356,8 +356,8 @@ class connection_args(ctypes.Structure):
 class access_rights_handler_args(ctypes.Structure):
     "access rights arguments"
     _fields_ = [('chid', chid_t),
-                ('read_access', ctypes.c_char),
-                ('write_access', ctypes.c_char)]
+                ('read_access', ctypes.c_bool),
+                ('write_access', ctypes.c_bool)]
 
 if PY64_WINDOWS and PY_MAJOR == 2:
     # need to add padding on 64-bit Windows for Python2 -- yuck!
--- a/epics/dbr.py
+++ b/epics/dbr.py
@@ -357,7 +357,7 @@ class connection_args(ctypes.Structure):
 class access_rights_handler_args(ctypes.Structure):
     "access rights arguments"
     _fields_ = [('chid', chid_t),
-                ('access', ctypes.c_ubyte)]
+                ('access', ubyte_t)]
 
 if PY64_WINDOWS and PY_MAJOR == 2:
     # need to add padding on 64-bit Windows for Python2 -- yuck!
@@ -382,5 +382,4 @@ if PY64_WINDOWS and PY_MAJOR == 2:
         "access rights arguments"
         _fields_ = [('chid', chid_t),
                     ('_pad_',ctypes.c_int8),
-                    ('read_access', uint_t, 1),
-                    ('write_access', uint_t, 1)]
+                    ('access', ubyte_t)]
--- a/epics/dbr.py
+++ b/epics/dbr.py
@@ -353,11 +353,11 @@ class connection_args(ctypes.Structure):
     _fields_ = [('chid', chid_t),
                 ('op', long_t)]
 
+
 class access_rights_handler_args(ctypes.Structure):
     "access rights arguments"
     _fields_ = [('chid', chid_t),
-                ('read_access', ctypes.c_bool),
-                ('write_access', ctypes.c_bool)]
+                ('access', ctypes.c_ubyte)]
 
 if PY64_WINDOWS and PY_MAJOR == 2:
     # need to add padding on 64-bit Windows for Python2 -- yuck!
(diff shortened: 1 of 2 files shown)


newville:
@klauer @dchabot I don't understand the `test_permit_enabled` failures in `test_cas.py`. 

This code works for me with Python 3.6.8, 3.7.5, and 3.7.6  to a) not fail on import, and b) correctly see "read-write" and "read-only" PVs.  

But I have not tested what happens if a connected PVs access changes state. I guess that is what this test is for, though I don't fully understand that database logic here. Could there be some race condition or timing issue with that `rps_lock` in the CAS tests?

epics/dbr.py
```diff
@@ -356,8 +356,8 @@ class connection_args(ctypes.Structure):
 class access_rights_handler_args(ctypes.Structure):
     "access rights arguments"
     _fields_ = [('chid', chid_t),
-                ('read_access', uint_t, 1),
-                ('write_access', uint_t, 1)]
+                ('read_access', ctypes.c_bool),
```
klauer: `c_bool` is wrong here, as I believe it's just an alias for `char`. 

That'd mean our previous bitfield which indicated `2 bits of 1 full byte` is now 2 full bytes instead.

newville:
@klauer @dchabot Thanks!  I was definitely just trying stuff.

I have since set up a softIoc using the scripts from test_cas to play with this some.   I may want to look into how to disable read access (though, I can' say I've ever seen that....).  

But, as you suggest, using
```
class access_rights_handler_args(ctypes.Structure):
    "access rights arguments"
    _fields_ = [('chid', chid_t), ('access', ctypes.c_ubyte)]
```

and
```
def _onAccessRightsEvent(args):
    'Access rights callback'
    try:
        entry = _chid_cache[_chid_to_int(args.chid)]
    except KeyError:
        return

    read = bool(args.access & 1)
    write = bool((args.access >> 1) & 1)
    entry.run_access_event_callbacks(read, write)
```

works.  And it works with both 3.7.5 and 3.7.6,  which is nice.   It seems to not matter whether it is "ubyte" or "uint", except for the number of higher order bits of garbage in the higher order bits.

I've pushed that..... fingers crossed.

newville:
@dchabot @klauer  OK, this looks like it passes with all Travis configs and works for me on Py 3.7.6 (Linux). I'm going to make another commit, to include this fix for the Win64+Python2 case.  That will pass because we don't actually test that combination. 

Is it OK with you 2 to then "Squash and Merge" and tag+release this as 3.4.1?

A couple other topics I'd like your opinions on:

  a) Should/Can we add Py 3.7.5 and 3.7.6 to the testing matrix?
  b) At what point can we drop Python 2 support?  To date it hasn't been much of a problem, but it would be nice to be able to assume Python 3.6+ soon.

klauer:
Sounds fine to me, @newville.

a.) I think 3.7-latest is good enough in the test matrix
b.) Perhaps make pyepics 3.x the last 2.7-compatible version, and bump to 4.0, removing the `if PY2`-style checks. Considering official support of 2.7 is over, I see zero issues doing this now. Anyone still using 2.7 can maintain their own version / hold back pyepics.

newville:
@klauer Thanks, I'll merge, tag, and push this out as 3.4.1, hopefully by tomorrow.

I'll also add Py3.7-latest to the test matrix.  

For Python 2.7 support, I mostly agree that anyone using Python2.7 is probably not keeping up-to-date with pyepics changes either.  I'll create a new issue with a proposed order of actions.

Issue #192, opened 2020-01-29
wx.BitmapButton : Additional wxPython Widgets for pyepics


vstadnytskyi:
I have found myself in a need to create a Button with a bitmap instead of a text. I want to make nice buttons with arrows on them. See the image below. 
<img width="75" alt="image" src="https://user-images.githubusercontent.com/29989124/73406858-20841280-42bd-11ea-8712-fd2ce90959cd.png">

Would it be useful to have PVBitmapButton next to PVButton in pyepics?

newville:
@vstadnytskyi  Sure, why not?   I have at least one app where I use custom arrows to move motors, but I guess I'm not using a PV mixin class for that.

vstadnytskyi:
A pull-request was created and added to the master branch. So I think we can close this issue.

PR #197, opened 2020-02-01
epics wx.PVBitmapButton

vstadnytskyi:
Addresses a proposed addition in the issue #192  This pull-request will add a new Mixin class epics.wx.PVBitmapButton. We can use bitmaps built-into wx.Python via wx.ArtProvider (https://wxpython.org/Phoenix/docs/html/wx.ArtProvider.html)

Here is an example:
image


vstadnytskyi:
We don't need to add .bmp images from this pull-request. I propose to you build in images from wx.Python wx.ArtProvider  library.

newville:
@vstadnytskyi it looks to me like this PR would include image files.  Is that the intention?

vstadnytskyi:
> @vstadnytskyi it looks to me like this PR would include image files. Is that the intention?

I have added them by a mistake. I am trying to figure out how to exclude them from the pull-request. sorry for the confusion.

newville:
@vstadnytskyi  in the interest of not losing this contribution, I'm going to merge this and then remove the image files.

Issue #198, opened 2020-03-26
Discrepancy between caput() behavior and documentation


daykin:
Hello,

I am not sure if this is the intended behavior, but if a PV does not exist, caput() returns None instead of a negative number like the documentation says.
Ex: 
```
import epics
print(caput("Brfxxccxxmnpcccclllmmnprxvclmnckssqlbb11116", 1, timeout=1))
```
output:
```
cannot connect to Brfxxccxxmnpcccclllmmnprxvclmnckssqlbb11116
None
```

newville:
fixed in master branch

PR #199, opened 2020-03-26
return -1 instead of None if caput times out on disconnected PV

daykin:
## Description

caputs to nonexistent/disconnected PVs currently return None instead of -1 like the docs suggest.

Issue created [here](https://github.com/pyepics/pyepics/issues/198).

### DIFF
--- a/epics/__init__.py
+++ b/epics/__init__.py
@@ -66,6 +66,8 @@ def caput(pvname, value, wait=False, timeout=60):
     if thispv.connected:
         timeout -= (time.time() - start_time)
         return thispv.put(value, wait=wait, timeout=timeout)
+    else:
+        return -1
 
 def caget(pvname, as_string=False, count=None, as_numpy=True,
           use_monitor=False, timeout=5.0):


newville:
@daykin  Maybe it actually makes more sense to return `None` here.  That is never going to be a valid value for a PV (whereas -1 can be).    That is, `None` is sort of a clearly non-valid value, almost an exception.   

I think we have had the conversation before about whether "not connected" should raise an exception, and it may be worth having that discussion again.   I think that the consensus is that "not connected" might mean "not connected yet" or "not connected right now" making it unclear whether a slightly unreliable network should cause an exception.  

So: maybe change the docs instead?

daykin:
I have no preference one way or the other in the -1 vs. None decision, and your justification makes plenty of sense. I can submit a PR with the docs changed instead.

My arguments for raising an exception: 
- Another implementation we also use at FRIB (cothread) does throw an exception
- Be it an unreliable network or nonexistent PV, we are unable to reach the PV in some allotted period of time. This is an unexpected event regardless of source, and as such, is worth raising an exception.

newville:
@daykin   Let's change the docs to reflect "return `None`"

Having `caget()` raise an exception instead of returning `None` would mean that any `caget()` outside of a try/except block could terminate a program if a PV is not currently connected.   There are lots of reasons that can cause that, and we don't want programs to terminate when an IOC gets rebooted or just an intermittent network loss.   

I don't know about you, but I sort of prefer that my browser does not exit on every momentary loss of wifi signal ;).

newville:
@daykin I'm going to close this and update the docs to say that caget returns `None` on an unconnected PV.

PR #200, opened 2020-04-07
Use a single connection timeout for all PVs in caget_many.

mattgibbs:
## Description
This PR makes `epics.caget_many()` much faster in instances where some of the PVs don't connect.  Instead of using `ca.connect_channel()` (which will connect to each channel serially, with a separate timeout for each), I've changed it to create all the channels, then loop over all of them, repeatedly checking connection status.  If all PVs aren't connected after a new `conn_timeout` parameter, they are skipped when getting values.

Previously, `caget_many()` would wait for up to `len(pvlist)*timeout`, which could be minutes if there were many disconnected PVs.  Now, it is possible to give up on those disconnected PVs much quicker.

While writing this PR I found that the implementation is really similar to something @brunoseivam mentions in #141.

### DIFF
--- a/epics/__init__.py
+++ b/epics/__init__.py
@@ -163,10 +163,6 @@ def caget_many(pvlist, as_string=False, count=None, as_numpy=True, timeout=5.0, 
     This does not maintain PV objects, and works as fast
     as possible to fetch many values.
     """
-    """get values for a list of PVs
-    This does not maintain PV objects, and works as fast
-    as possible to fetch many values.
-    """
     chids, out = [], []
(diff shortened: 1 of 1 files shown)


newville:
@mattgibbs Oh, sorry about that.  I think that I thought that we had solved this back then, but clearly we did not.     So, thanks!

I'm going to merge this but then probably rewrite it some.   I don't know if I can make it faster, but I will try to not make it slow for "large number of unconnected PVs".    But it also reminds me that we should really be working on the testing mechanism so that we can better test this sort of thing.

PR #201, opened 2020-04-08
caget_many(): cleanups (and maybe speed ups, and better timing accuracy)

newville:
This is a slight cleanup of #200, partly addressing #141, and should give better timing accuracy.  

I tested with 10000 PVs, 95 of which are not connected and the rest scalar values on a single subnet.

Using `conn_timeout=1.0` (now the default), I see the runtime drop from about 2.4 seconds (current master, after #200) to about 1.6 seconds (this PR), while also consistently getting exactly 95 `None` values.


newville:
OK, I'm going to merge this.  I think it is uncontroversial....

PR #203, opened 2020-05-08
FIX: do not block if disconnected when changing auto_monitor settings

klauer:
## Description
Calling `clear_auto_monitor` when an `epics.PV` is disconnected may cause it to attempt to reconnect when accessing the `count` property.

Detailed in the following traceback:

<details>

```python
Traceback (most recent call last):
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/weakref.py", line 624, in _exitfunc
    f()
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/weakref.py", line 548, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/ophyd/signal.py", line 1718, in destroy
    self.cl.release_pvs(self._write_pv)
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/ophyd/_pyepics_shim.py", line 122, in release_pvs
    pv.clear_auto_monitor()
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 48, in wrapped
    return func(self, *args, **kwargs)
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 444, in clear_auto_monitor
    self.auto_monitor = False
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 48, in wrapped
    return func(self, *args, **kwargs)
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 409, in auto_monitor
    self._check_auto_monitor()
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 48, in wrapped
    return func(self, *args, **kwargs)
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 357, in _check_auto_monitor
    count = self.count
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/epics/pv.py", line 962, in count
    return self._getarg('count')
  File "/Users/klauer/mc/envs/lucid/lib/python3.6/site-packages/ophyd/_pyepics_shim.py", line 92, in _getarg
    self.get_ctrlvars(timeout=1, warn=False)
[...]

PR #204, opened 2020-05-15
FIX: Do not block if disconnected when changing auto_monitor settings - continued

hhslepicka:
## Description
This PR is a continuation of the work started by @klauer at https://github.com/pyepics/pyepics/pull/203.

The tests were failing because `_check_auto_monitor` was never being called after `self.connected` was set. Moving it after that makes the tests pass locally and also fixes the tests hanging at Typhos.

🤞 that they will pass at Travis here as well.

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -333,7 +333,8 @@ class PV(object):
         # threads from thinking a connection is complete when it is actually
         # still in progress.
         self.connected = conn
-        self._check_auto_monitor()
+        if conn:
+            self._check_auto_monitor()
 
     @_ensure_context
     def _clear_auto_monitor_subscription(self):
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -319,7 +319,6 @@ class PV(object):
             self._args['type'] = _ftype_
             self._args['typefull'] = _ftype_
             self._args['ftype'] = dbr.Name(_ftype_, reverse=True)
-            self._check_auto_monitor()
 
         for conn_cb in self.connection_callbacks:
             if callable(conn_cb):
@@ -334,6 +333,7 @@ class PV(object):
         # threads from thinking a connection is complete when it is actually
         # still in progress.
         self.connected = conn
+        self._check_auto_monitor()
 
     @_ensure_context
     def _clear_auto_monitor_subscription(self):
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -354,10 +354,13 @@ class PV(object):
 
         Clears or adds monitor, if necessary.
         '''
-        count = self.count
-        chid = self.chid
+        if not self.connected or self.chid is None:
+            # Auto-monitor will be enabled (or toggled based on count) upon the
+            # next connection callback.
+            return
 
-        if count is None or chid is None:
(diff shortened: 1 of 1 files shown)


newville:
@hhslepicka @klauer This change is passing, so I guess "maybe", but I'm confused by this.  

Doesn't moving that call of`self._check_auto_monitor()` to the outside of the `if conn:` block mean that this will also be called on disconnect events?  

Won't that explicitly clear and then re-create the ca_subscription on disconnect events?  That seems not-quite-optimal to me, but maybe I'm missing something.  

I think it is ok to call the users' connection callbacks and then create the subscription, but we should think about the possibility of running two successive connection callbacks.   I don't think that will happen, but I've been wrong before ;).

What is the issue that causes the problem?

Anyway, I'm not really opposed to this, just mildly concerned about messing with dragons.

hhslepicka:
@newville the very first check at the _check_auto_monitor will bail out if not connected or if chid is None so it should not be a problem.

newville:
@hhslepicka then why even call the function if not connected?

klauer:
I think the key is: https://github.com/pyepics/pyepics/pull/204/commits/d9e199e9dca74278ed6476660771070427cf4c81#diff-e10bca7150af266943a84e8a4233a349R335

`self.connected = True` allows the auto-monitor to be configured, whereas in the previous location (my commit) it was not yet set.

newville:
@klauer yeah, but in the current master, wasn't `_check_auto_monitor()` being called only on connection events?    I guess maybe a problem could arise if one accesses the `auto_monitor` property of an unconnected PV, but other than that, I'm having a hard time seeing how you get this sort of problem.

It definitely seems OK to check the connection status at the top of `_check_auto_monitor()`.  But in `_on_connect()`, where you know the connection state just changed, why even bother calling it if `conn` is `False`?  So maybe just make that::
 
    self.connected = conn 
    if conn:  
       self._check_auto_monitor()

Other than that, this appears to move the calling of the user's connection callbacks to be before the actual subscriptions are made.   I don't know if that a big deal - I don't think we rely on that order.

klauer:
Yeah, true - the `if conn` would not be unreasonable there, even if it duplicates the check in `_check_auto_monitor`.

I think the ordering of: "get connection callback, start value monitoring" is actually better than the other way around (current master).

We ran into this bug during teardown and cleanup. For whatever reason, if the PV is disconnected `clear_auto_monitor()` shouldn't try to reconnect - which is what this PR is after. I think we can agree on that point.

hhslepicka:
@newville I pushed the requested change to wrap it with a check for `conn`.

newville:
@klauer @hhslepicka yep, I think that all makes sense, and thanks for making that change.  This looks ok to merge.  Any objections or concerns from anyone else?  If you want to merge this now that's fine me.  If not, I'll plan to merge in a day or so.

klauer:
Since we're all good with it (and I think it'll stop some of our downstream library issues), I think we can merge if @hhslepicka is ready.

hhslepicka:
I am good to go if y'all are! 👍

PR #205, opened 2020-05-28
Prevent exceptions after libca is finalized during teardown

MikeHart85:
## Description

This PR builds on @klauer's suggested fix for an issue that began in https://github.com/bluesky/ophyd/issues/834 and resurfaced in https://github.com/bluesky/ophyd/pull/866. This also affected several downstream projects, which began seeing unexpected exceptions in their test suites.

The issue occurs when Python is shutting down. Since the order in which Python tears everything down is not very consistent or controllable, libca is sometimes finalized before downstream code has a chance to run its cleanup. When that cleanup later invokes pyepics functionality that requires libca, exceptions are thrown. Some of them cannot be caught.

An example of this can be triggered by running the following snippet against the current master:

```python
import epics

class Broken:
    def __init__(self, pv):
        self.pv = epics.get_pv(pv)
    def __del__(self):
        self.pv.clear_auto_monitor()

mypv = Broken('sim:mtr1')
``` 

This causes a `ChannelAccessException`, which can be caught. But also prints `ImportError` and `KeyError` tracebacks, which cannot be caught.

Ken's commit replaces this behavior with a single RuntimeError, raised in `@withCA`, which can be caught.

My commit modifies this to:

1. Simply `return` instead of raising a RuntimeError
2. Also do the same in the various `@with...CHID` decorators

This means that any function using `@withCA` or any of the CHID decorators will silently do nothing and return `None` if `finalize_libca` has been called. I realize this is an aggressive thing to do, and it would be fair to argue that client code should just `try` and catch the exceptions instead. My rationale for returning instead is as follows:

If `finalize_libca` has been called, that means `atexit` has been triggered. This means Python is shutting down, and it happened to decide to clean up pyepics before the client code has finished finalizing. There is no way to recover from this; at best client code can ignore the exception and carry on. There's no benefit to forcing client code to wrap all pyepics usage, in any code branch that might be invoked during teardown, in `try: ... except: pass` boilerplate (even functions that don't raise anything under any other circumstance). Philosophically speaking, code using pyepics shouldn't have to worry about an internal dependency of pyepics going away before they were done using pyepics.

The reason for 2. is that many functions only use the CHID decorators (without `@withCA`). And so, for example, calling `epics.ca.clear_channel()` after libca is finalized results in a `ChannelAccessException: Unexpected channel ID` even though the channel ID was previously valid and not yet cleared by client code. This is because `finalize_libca` clears `_chid_cache`.

Whether exception or silent `return`, I think it makes sense to have the same failure behavior from all functions when called after libca is finalized.

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -545,9 +545,7 @@ def withCA(fcn):
         if libca is None:
             initialize_libca()
         elif libca is _LIBCA_FINALIZED:
-            # See what happens if you uncomment this instead:
-            # initialize_libca()
-            raise RuntimeError('libca shutting down')
+            return  # Avoid raising exceptions when Python shutting down
         return fcn(*args, **kwds)
     return wrapper
 
@@ -563,6 +561,10 @@ def withCHID(fcn):
     @functools.wraps(fcn)
     def wrapper(*args, **kwds):
         "withCHID wrapper"
+        global libca
+        if libca is _LIBCA_FINALIZED:
+            return  # Avoid raising exceptions when Python shutting down
+
         if len(args)>0:
             chid = args[0]
             args = list(args)
@@ -587,6 +589,10 @@ def withConnectedCHID(fcn):
     @functools.wraps(fcn)
     def wrapper(*args, **kwds):
         "withConnectedCHID wrapper"
+        global libca
+        if libca is _LIBCA_FINALIZED:
+            return  # Avoid raising exceptions when Python shutting down
+
         if len(args)>0:
             chid = args[0]
             args = list(args)
@@ -613,6 +619,10 @@ def withMaybeConnectedCHID(fcn):
     @functools.wraps(fcn)
     def wrapper(*args, **kwds):
         "withMaybeConnectedCHID wrapper"
+        global libca
+        if libca is _LIBCA_FINALIZED:
+            return  # Avoid raising exceptions when Python shutting down
+
         if len(args)>0:
             chid = args[0]
             args = list(args)
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -65,6 +65,9 @@ PREEMPTIVE_CALLBACK = True
 
 AUTO_CLEANUP = True
 
+# A sentinel to mark libca as going through the shutdown process
+_LIBCA_FINALIZED = object()
+
 ##
 # maximum element count for auto-monitoring of PVs in epics.pv
 # and for automatic conversion of numerical array data to numpy arrays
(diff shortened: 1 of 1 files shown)


klauer:
As a partial contributor to this (and one affected by the shutdown mess), I'm overall in support of this.

If there is any legitimate reason one would want to call `finalize_libca` during the lifetime of their application and not just at teardown (I personally can't think of one), there could be two sentinels introduced: call `finalize_libca()` -> set sentinel to `_user_requested_finalize` -> if user calls CA functions again `raise RuntimeError('You called finalize_libca; please re-initialize it.')`. The `atexit` sentinel could be switched to something like `_atexit_finalize`, which would _not_ raise during these calls (as in the current PR).

To be clear: I don't think the above is really necessary. I think this PR is sufficient, unless alternative `finalize_libca` use cases can be presented.

newville:
@MikeHart85 @klauer Thanks - this seems fine to me.  

That is, I think it is fine to signal that `finalize_libca()` has been called, and avoid doing any deliberate communication with CA after that until the end of the process".   I agree that there is no real use case (or, well, no current support) for wanting to call `finalize_libca()` and then re-initializing libca.  If someone needs that at some point, we can try to deal with it then.  

My only question / vague-concern is : should the global `libca` be set to a sentinel value or should there just be a global "LIBCA_FINALIZING" (or something) that starts as `False` then get sets to `True` in `finalize_libca()`, with checks being like `if libca is None or LIBCA_FINALIZING:`? 

I don't have a strong opinion, but is making the pointer to the library held in `libca` completely unreachable an important part of this fix, helping prevent the underlying issues?

MikeHart85:
It should be possible to achieve the same thing with a separate state variable.

The `libca` global was reset to `None` in `finalize_libca` prior to this change, so it would be unreachable after finalizing in any case. The only (minor) difference I can think of is that this way you don't have to remember to set both `libca` and some other state variable every time the state changes, avoiding situations where they accidentally go out of sync. It might also be marginally more thread safe for much the same reason? Not sure if that would ever come into play.

I don't have any strong preference either though and would happy to change it if you prefer?

newville:
@MikeHart85 @klauer Yep, that all looks good!  Thanks!

prjemian:
@newville : When can we expect a release and a conda package with this revision?

hhslepicka:
@prjemian the release was made... the conda-package will be out soon via conda-forge.

prjemian:
Thanks.  Missed that release by mere minutes, it seemed.  About 10 of them, to be precise.

hhslepicka:
As soon as tests are passing at the feedstock I will merge it in: https://github.com/conda-forge/pyepics-feedstock/pull/23

After that it will take about 15-20 minutes for it to be available.

PR #209, opened 2020-06-22
FIX: Clean access and connection callbacks at clear_callbacks.

### LINKED ISSUE #208: clear_callbacks leaves connection and access callbacks alive
Users reported some issues with PyDM recently that when IOCs are rebooted they get a message that connection and access control callbacks failed even if the widget was no longer at the screen.
I tracked it down and I was able to isolate the issue into an example:

Instructions to reproduce:
- Run the code below...
- When the code goes into the poll loop, kill the IOC
- Note on the terminal that even with the PV disconnected the connection and access callbacks are invoked.

**Note:** Change the PV from `MTEST:Float` to something else that you have.
```python
import epics
import time


def conn_callback(*args, **kwargs):
    print(f'{time.time()} conn_callback invoked with: {args} | {kwargs}')


def access_callback(*args, **kwargs):
    print(f'{time.time()} access_callback invoked with: {args} | {kwargs}')


print(f'{time.time()} - Create PV')
pv = epics.PV('MTEST:Float', connection_callback=conn_callback, access_callback=access_callback)

print(f'{time.time()} - Connect')
pv.connect()
print(f'{time.time()} - Wait For Connection')
pv.wait_for_connection()

print(f'{time.time()} - Invoke disconnect')
pv.disconnect()

print(f'{time.time()} - Lets enter on poll loop')
try:
    while True:
        epics.ca.poll()
except KeyboardInterrupt:
    print(f'{time.time()} - Poll loop terminated')
    pass
```

This produces the following output:
```
1592863195.09071 - Create PV
1592863195.111618 - Connect
1592863195.142028 access_callback invoked with: (True, True) | {'pv': <PV 'MTEST:Float': not connected>}
1592863195.142275 conn_callback invoked with: () | {'pvname': 'MTEST:Float', 'conn': True, 'pv': <PV 'MTEST:Float': not connected>}
1592863195.14254 - Wait For Connection
1592863195.1425579 - Invoke disconnect
1592863195.144663 - Lets enter on poll loop


CA.Client.Exception...............................................
    Warning: "Virtual circuit disconnect"
    Context: "10.0.0.211:5064"
    Source File: ../cac.cpp line 1237
    Current Time: Mon Jun 22 2020 15:00:04.489395534
..................................................................
1592863204.4895098 conn_callback invoked with: () | {'pvname': 'MTEST:Float', 'conn': False, 'pv': <PV 'MTEST:Float': not connected>}
1592863204.489681 access_callback invoked with: (False, False) | {'pv': <PV 'MTEST:Float': not connected>}
```

For now I can force clear the connection and access callbacks on my side.

I would like to pick your brain before submitting a PR as what would be the best fix for this case:
- Add the cleaning routine at `clear_callbacks` to also clear the other two callbacks
- Other alternatives?

Could this also be related to https://github.com/pyepics/pyepics/issues/156 ?

hhslepicka:
## Description
Closes #208

### DIFF
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -831,11 +831,13 @@ class PV(object):
             self.callbacks.pop(index)
             ca.poll()
 
-    def clear_callbacks(self):
+    def clear_callbacks(self, with_access_callback=False, with_connect_callback=False):
         "clear all callbacks"
         self.callbacks.clear()
-        self.access_callbacks = []
-        self.connection_callbacks = []
+        if with_access_callback:
+            self.access_callbacks = []
+        if with_connect_callback:
+            self.connection_callbacks = []
 
     def _getinfo(self):
         "get information paragraph"
@@ -1118,7 +1120,7 @@ class PV(object):
 
         self._monref = None
         self._monref_mask = None
-        self.clear_callbacks()
+        self.clear_callbacks(True, True)
         self._args = {}.fromkeys(self._fields)
         ca.poll(evt=1.e-3, iot=1.0)
 
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -834,6 +834,8 @@ class PV(object):
     def clear_callbacks(self):
         "clear all callbacks"
         self.callbacks.clear()
+        self.access_callbacks.clear()
+        self.connection_callbacks.clear()
 
     def _getinfo(self):
         "get information paragraph"
--- a/epics/pv.py
+++ b/epics/pv.py
@@ -834,8 +834,8 @@ class PV(object):
     def clear_callbacks(self):
         "clear all callbacks"
         self.callbacks.clear()
-        self.access_callbacks.clear()
-        self.connection_callbacks.clear()
+        self.access_callbacks = []
+        self.connection_callbacks = []
 
     def _getinfo(self):
         "get information paragraph"


klauer:
Very reasonable that `clear_callbacks` would clear _all_ of them, not just value callbacks. 

While this is a small API break, it seems the correct path to me. 👍

newville:
Maybe add a `with_access_callback` and  `with_connect_callback` arguments to `clear_callbacks()`?
Maybe `PV.disconnect()` should just explicitly clear these?

hhslepicka:
@newville is your suggestion to move the cleanup of connect_cb and access_cb out of clear_callbacks and into disconnect? Or to add options to clear_callbacks with default being False for those to be cleaned up and pass them as True for disconnect ?
Is that the case?

newville:
@hhslepicka @klauer Just as an aside:  I think we have had no policy on this, but I would be willing to drop Python2 support at any time for the development branch and all future releases.   I'm sure some pyepics users are stuck with Py2, but not having access to any changes we make from here on is probably the least of their problems ;).

newville:
@hhslepicka  They were alternate suggestions to consider.   I don't have a strong preference for which of these is adopted.

hhslepicka:
@newville I pushed a commit with the suggested changes. This reverts a regular call to `clear_callbacks` to work as before and `disconnect` explicitly calls it with True for both callbacks to be cleared as well.

newville:
@hhslepicka Thanks!  I think having the arguments to `clear_callbacks()` is probably a better approach, but I'm happy to hear other suggestions or preferences.

newville:
@klauer @hhslepicka  This all looks good to me.... any objections to merging?

hhslepicka:
@newville no objections for me.

newville:
@hhslepicka  @klauer Thanks, I'll merge this!

Issue #213, opened 2020-09-17
pyepics version 3.5.0: end of life for Python2.7 and Python 3.5


newville:
@klauer @hhslepicka.  I plan to push a minor release (3.4.3) from the minor changes made since 3.4.2. 

I propose that this be the final version to support Python 2.7 or 3.5, and that we plan to release a pyepics 3.5.0 soon-ish (say, November 1) that has no real feature changes from 3.4.3, but requires Python 3.6, and distributes shared libs built with epics base 7.0.4.

I would like to review and maybe revise the testing setup too, but I'm not certain that will change.

Any objections to this?

klauer:
No objections from me.

hhslepicka:
I am good with that! 👍

Issue #212, opened 2020-09-17
PV status/severity -- potential error in the docs?


NichtJens:
I was under the impression that `pv.status` and `pv.severity` are equivalent to the alarms defined in:
https://git.launchpad.net/epics-base/tree/src/libCom/misc/alarmString.c?h=3.15

However, the docs say:

>`pv.status`
>    The PV status, which will be **1** for a Normal, connected PV.

and

> _status_: the status of the PV (**1** for OK)

If my reading of the alarm strings is correct, the status for a normal and successfully connected PV is **0** there. The results of `pv.status` I am seeing, also seem to be consistent with the alarm strings and not with the pyepics docs.

newville:
@NichtJens Oh thanks, yes I agree the doc is wrong there.   Will fix.

newville:
@NichtJens Fixed in master branch.  I will push a new minor release (3.4.3) and publish docs soon-ish (hopefully today).

NichtJens:
@newville Great!

It might be worth showing the conversion table from status number to string in the docs as well. Or simply pointing to `alarmString.c`. Even better would be an actual dict / conversion function. Would you be interested in a PR for the latter?

newville:
@NichtJens yes, I agree that status/severity string messages would be very nice.  A PR for that would be most welcome!

klauer:
If we're targeting only Python 3 now, `enum.IntEnum` can clean these up nicely while providing both integer values and convenient user-facing representations (`AlarmStatus.NO_ALARM` <-> `0`).

newville:
@klauer +1 from me - I think going Py3.6+ would allow a lot of cleanups and modernization.  (and I guess that's a reason to leave the tests alone for a while).

NichtJens:
If I am not mistaken, `enum` only nicely allows this direction: `AlarmStatus.NO_ALARM` -> `0` ...

But how do you use them for this direction: `0` -> `AlarmStatus.NO_ALARM` ?

I would think the former case is not really needed here, though. We _have_ `pv.status` and _want_ a human-readable string (or enum entry name, if we go with those). I don't see how we _have_ the string and _want_ the error code.

I would have gone with a dict `{0: "NO_ALARM", ...}`. It could even be a list/tuple, but then the indices are not as explicit, which might cause problems if -- however unlikely -- a code gets removed from the middle of the range.

But maybe I am missing something obvious.

klauer:
```python
[...]

PR #214, opened 2020-09-18
Human-readable alarm status and severity for PVs

NichtJens:
## Description

As discussed in #212, this PR adds `PV.char_status` and `PV.char_severity` as human-readable companion to `PV.status` and `PV.severity`. Both are also shown in `PV.info`.

In the back, this uses two enums (`AlarmStatus`, `AlarmSeverity`) via a custom `DefaultIntEnum` (`enum.IntEnum` with fake default member "UNKNOWN" for missing entries).

[To be merged after moving pyepics to support only Python 3.6+ (cf. #213)]


newville:
@NichtJens can you please rebase so that changes to `dbr.py` from #215 are reflected here.  The good news is that by testing only with Py36,  your changes here should pass the tests and we will merge them in.   Thanks!

NichtJens:
@newville Hi! Sorry for the slight delay. Seems I rebased correctly now, and the tests indeed pass.

newville:
@NichtJens thanks!

PR #215, opened 2020-10-08
Drop Python2 and 3.5 (and iron python)

newville:
## Description

drop support for Python 2.7 and 3.5.


newville:
I would like to merge this soon-ish (next couple of days) so that we can then merge in #214 (which might need a rebase for `dbr.py`).    
This may not be done and/or need tweaking, but I'd like to get this into master before making more aggressive changes to the testing setup.

epics/ca.py
```diff
@@ -26,10 +26,10 @@
 
 from math import log10
 from pkg_resources import resource_filename
+from copy import deepcopy
 
-from .utils import (STR2BYTES, BYTES2STR, NULLCHAR_2,
-                    strjoin, memcopy, is_string, is_string_or_bytes,
-                    ascii_string, clib_search_path)
+from .utils import (str2bytes, bytes2str, strjoin,
```
klauer, line 31: Name changes are a step in the right direction, but you may consider back-compat for the old names. Even if you don't consider these part of the public API, I'd bet that some out there are using it:

```
# Back-compat; deprecated and to be dropped in pyepics x.y.z
STR2BYTES = str2bytes
```

klauer (commented):
Seems good overall 👍 

Farewell, Python 2!

epics/ca.py
newville: @klauer   Yeah, I think you're probably right, I will add back-compatible aliases for all of those legacy names in `utils.py` where they can enjoy a long and boring life now that we know we don't really need them.

newville:
ok, merging this...

PR #219, opened 2021-05-13
PycSimpleType appearing in the function native_type

hanak:
## Description

**Problem**
When a get callback receives a type it does not recognize, it will raise TypeError due to invalid use of type. It is caused by a fallback type instance which not an integer as the rest of the code expects.

**Solution**
1) This update fails casting if the type is not recognized because it is not able to tell much about the supplied memory pointer. In case of `_onGetEvent`, this will yield an error. In case of `_onMonitorEvent` this will end the function silently similar to other cases here.
2) This update adds a unit test which covers the behavior `bdr.cast_args`. This unit test is included into the tests in `.travis.yml`.
3) This update adds `venv` and by-products of vscode and `pip install -e .` into `.gitignore'.

**Test**
1) `test_bdr` was run manually.
2) Updated pyepics was used to retrieve a PV from a running IOC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,7 @@ tests/.coverage*
 build
 dist
 NonGit
+venv
+.vscode
+pyepics.egg-info
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,16 @@ script:
   - caget Py:long1
 
   # running tests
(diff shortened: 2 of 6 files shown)


newville:
@hanak OK, thanks.  I'll merge this -- it mostly reminds me that I need to fix the CI testing

Issue #156, opened 2019-04-30
ca.clear_channel should be called on PV.disconnect


klauer:
Current master (and all previous released versions) appear to leave a CA channel dangling without reference in Python - until the interpreter exits.

Need to test that adding this causes no new issues or instability.

juanfem:
Hello. Is there any reason why this hasn't been added yet?

At ESS we are struggling with memory leaks that we think are caused by this. Adding a call to ca.clear_channel() after PV.disconnect() seems to solve the issue for us.

juanfem:
Hi again,

I think I understand the issue now.

I assume that ca.clear_channel() is not called because there can be more than one PV object connecting to the same channel and pyepics is caching that connection. Therefore, PV.disconnect() should only call ca.clear_channel() when there are no PV objects using it.

However, [here](https://github.com/pyepics/pyepics/blob/d39904efc46a8573366007a6576ff43a634f1854/epics/pv.py#L1121) the reference to that channel is removed from ca._cache by calling ca._cache.pop().

I think a solution could be to keep a list of all the PV objects subscribing to the same channel and only remove the channel from ca._cache and call ca.clear_channel() when the las PV object calls PV.disconnect().

I can volunteer to create a pull request if needed.

Cheers,
Juan

newville:
@juanfem Before working on a solution, it would be helpful to understand the problem you are seeing.  When you say "memory leaks" do you mean that memory usage keeps growing even though no resources (say, PVs) are explicitly allocated, or do you mean that trying to purge PVs from a program may leave some dangling connections?  

Diagnosing memory issues for Python programs can be challenging.  Diagnosing PVs being left partially in the `pyepics` caches ought to be easier to diagnose.   For sure, I and many other people run long-running pyepics-heavy applications.   I'm skeptical that there are "real" memory leaks, but I'll also admit that most of my long-running pyepics apps will connect to a set of PVs (perhaps only 25, perhaps a few thousand) and don't try to disconnect from PVs very often -- a Display Manager may very well want to connect/disconnect from PVs.

So, can you describe the actual problem you are seeing before trying to change the code?

emanuelelaface:
Hi, I work with @juanfem and I can post here the code. The main idea is to have something that returns all the details of a list of PVs in a FastAPI application. So the function is this:

```
@app.get("/api/v1/caget/{pvlist}")
async def get_pv(pvlist: str):
    pvnames = pvlist.split(",")
    pvs = []
[...]

PR #220, opened 2021-06-02
FIX memory leak on PV.disconnect() #156

juanfem:
This PR is a proposal to fix the issue #156 .

The issue occured because the disconnect method left some references to the PV object dangling, which prevented the garbage collector to collect the PV object.

The method behaves otherwise the same as before, i.e. the connection is still cached by epics.ca and reused if a new PV object is created for the same signal.

I also included a set of tests to verify the interaction of the disconnect method with other methods provided by the library.

### DIFF
--- a/.github/workflows/test-with-conda.yml
+++ b/.github/workflows/test-with-conda.yml
@@ -35,6 +35,7 @@ jobs:
         python setup.py install
     - name: Run test suite
       run: |
+        python -m pip install psutil
         cd tests/Setup
         export EPICS_CA_ADDR_LIST=localhost
         export EPICS_CA_AUTO_ADDR_LIST=NO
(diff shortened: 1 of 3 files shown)


newville:
@juanfem @klauer  Thanks again for tackling this.  I read through this again, verified it on one test machine (Linux) and am happy to see it passing tests without changing any existing tests.  

This method of using "count" and "remove" methods sure **looks** like the right sort of solution to me.   I think my main concern would be that `list.count()` and `list.remove()` are for cases where there may be more than one of repeat of the same item (ie, count > 1).  But in that case `list.remove()` removes the first one in the list.

Should this explicitly remove all of them, maybe with (for example):

           while cache_item.callbacks.count(self.__on_connect):
                cache_item.callbacks.remove(self.__on_connect)

I'm not sure that I have a strong preference (being always a little unsure of how complete `PV.disconnect()` really should be), but it does seem like if the count is > 2 that a single `remove` will still leave some methods
that can no longer be reached or garbage collected.

Any thoughts on that?  Am I missing something (for sure: that is a definite possibility!!).   Anyway, I am inclined to merge this unless there are objections or concerns raised.

newville:
@juanfem  With no comments or objections, my inclination is to merge this as is, and then create a new issue to ask whether that `List.remove()` should be "remove all".

Thanks!  I think this goes a long way to solving one of the longest-standing issues.

Issue #224, opened 2021-10-02
Missing runtime dependency: pkg_resources


eddybl:
I noticed that I cannot successfully import pyepics if I do not have setuptools / pkg_resources installed, although it does not seem to be declared as a runtime dependency:
```python
Traceback (most recent call last):
  File "<stdin>", line 20, in <module>
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/epics/__init__.py", line 27, in <module>
    from . import ca
  File "/usr/lib/python3/dist-packages/epics/ca.py", line 28, in <module>
    from pkg_resources import resource_filename
ModuleNotFoundError: No module named 'pkg_resources'
```

I assume this does not show up usually, since setuptools is typically installed in most cases. I was running into this issue inside a docker environment in which I installed pyepics (but where nothing else is installed unless it is a dependency)

PR #225, opened 2021-10-02
Add setuptools as runtime requirement

smartsammler:
setuptools' pkg_resources is used in the ca module, but setuptools was not
 listed as a dependency, yet.

## Description
Add setuptools as a dependency, because pkg_resources is used in epics/ca.py.
This tries to fix issue #224 
Though setuptools is installed most of the time, [pkg_resources does not ship with Python](https://www.python.org/dev/peps/pep-0365/) itself and hence shouldn't been taken for granted.


newville:
@smartsammler @eddybl Thanks!

Issue #228, opened 2021-11-02
pyepics crashes process on shutdown randomly


hanak:
**Problem**
I am using pyepics with behave in order to run BDD on IOCs. I am experiencing random crashes on the shutdown. There, libca reports "CA client library tcp receive thread terminating due to a non-standard C++ exception" and terminates behave. This is detected as test failure.

I was not able to find the cause, yet. Unfortunately, there is an exception trap in libca which logs the information and consumes the exception. This prevents the exception to bubble-up into the OS so that a debugger might get is easily. Yet, recently I run an instance which offered more information about potential cause:
```
Exception ignored on calling ctypes callback function: <function _onMonitorEvent at 0x7f2f72772160>
Traceback (most recent call last):
  File ".../site-packages/epics/ca.py", line 682, in _onMonitorEvent
Exception ignored on calling ctypes callback function: <function _onMonitorEvent at 0x7f2f72772160>
Traceback (most recent call last):
  File ".../site-packages/epics/ca.py", line 682, in _onMonitorEvent
    entry = _get_cache_by_chid(args.chid)
  File ".../site-packages/epics/ca.py", line 472, in _get_cache_by_chid
    entry = _get_cache_by_chid(args.chid)
  File ".../site-packages/epics/ca.py", line 472, in _get_cache_by_chid
    pvname = BYTES2STR(libca.ca_name(dbr.chid_t(chid)))
AttributeError: 'object' object has no attribute 'ca_name'
CA client library tcp receive thread terminating due to a non-standard C++ exception
FATAL: exception not rethrown
```

**Early Analysis**
Looking at `ca.py`, the variable `libca` is set to `_LIBCA_FINALIZED` on shutdown. `_LIBCA_FINALIZED` is `object`. So, in theory I might be looking at concurrency issue in pyepics: `_onMonitorEvent()` arrives while `libca` is set or being set to `_LIBCA_FINALIZED`.  Since `_onMonitorEvent()` (`ca.py`) does not expect it, it tries to use `libca` as if it was a valid object. This is suspicious but it could be just fine depending on intended use. Is this expected?

[...]

PR #229, opened 2021-11-26
clear_cache unregisters cached channels clearing

hanak:
## Problem
Using `ca.clear_cache()` may trigger a crash on shutdown. This happens randomly. Yet, it is more frequent if PVs in the cache update frequently. This behavior has been observed on Linux most of the time.

## Analysis
`ca.clear_cache()` removes reference to cached channels but it does not remove registration of those channels in EPICS. As a consequence, if timing is right, a callback may by invoked while libca is unloaded/detached already.

## Issue
https://github.com/pyepics/pyepics/issues/228

## Proposed Solution
Remove callback registration of cached channels prior removing the cache. This change is covered by a new automated test `test_ca_clearcache.py`. This test is quite slow because it tries to trigger a random concurrency issue. Furthermore, this update adds a protection from running `start_iocs.sh` from invalid location.

## Performed Tests
Use OS: Python 3.9, Linux 64-bit (Debian), Virtual Machine w/ 2 CPUs
* Run `softIoc ./st.cmd` followed by `pytest test_ca_clearcache.py`. Test passed.
* Run `tests/run_coverage.sh`. All tests passed.

## Known Limitations
* Proposed `ca.clear_cache()` set a property `chid` of cached channels to None prior clearing the cache. Previously, this was not done. If a user of pyepics keeps a reference to an instance of a channel (create through `ca.create_channel()`) after calling `ca.clear_cache()`, this might cause malfunction. However, I did not perform any deeper analysis.

## Questions
1) While updating various bash scripts I have noticed a file `tests/Coverage_Test.sh`. This file references .sh files which are not present. It looks like an obsolete version of `tests/run_coverage.sh`. Is this so? May I remove the file?


newville:
@hanak Thanks, and Sorry for the delay.  I think this looks fine, especially as it only affects `ca.clear_cache` (and is clearly an improvement there!).   

And, yes, it would be fine to remove `Coverage_Test.sh`.  If you do that here, I'll be happy to merge.

hanak:
@newville Thank you for the review. I have removed the file `Coverage_Test.sh`. Also, I have added test by-products (`tmp_aostate.txt` and `cage.tst`) to `.gitignore`.

newville:
@hanak Great! Thanks very much for that, including adding to `.gitignore`.  I think this will make `clear_cache` a much more usable method.

Issue #233, opened 2022-02-08
pyc files not removed properly from distributions


smartsammler:
`*.pyc` caching files are only removed from the root directory, but not recursively.

We are facing the problem that [stdeb](https://github.com/astraw/stdeb) cannot cope with .pyc files being part of the sdist tar.gz.
The newest release on PyPI (https://pypi.org/project/pyepics/#files [3.5.1](https://files.pythonhosted.org/packages/6c/cd/f97b676b6903e6f88bfced3817964559c20a1e3c6cd5244fa188f76fff75/pyepics-3.5.1.tar.gz)) includes such a file: `scripts/__pycache__/mpanel.cpython-37.pyc`, thus causing the mentioned issue with stdeb.

As the pyc files are already excluded from the main directory, I guess it is intended that these files are not part of the sdist—which is afaik best practice anyway.

I can provide a PR if appreciated.

PR #234, opened 2022-02-08
Exclude pyc files recursively

smartsammler:
## Description
Recursively exclude .pyc caching files from being included in source distributions in order to prevent issues with further build tools (like stdeb) or different setups trying to use the cache file though it might not match the current environment.

We were facing the problem that [stdeb](https://github.com/astraw/stdeb) cannot cope with .pyc files being part of the sdist tar.gz.
The newest release on PyPI (https://pypi.org/project/pyepics/#files 3.5.1 https://files.pythonhosted.org/packages/6c/cd/f97b676b6903e6f88bfced3817964559c20a1e3c6cd5244fa188f76fff75/pyepics-3.5.1.tar.gz) includes a `scripts/__pycache__/mpanel.cpython-37.pyc` file, thus causing the issue with stdeb.

As the pyc files are already excluded from the main directory, I guess it is intended, that these files are not part of the sdist—which is afaik best practice anyway.

This fixes [issue #233](https://github.com/pyepics/pyepics/issues/233)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,4 +9,4 @@ recursive-include conda-recipe *
 recursive-exclude doc/_build *
 recursive-exclude doc *.pdf
 recursive-exclude tests/Misc *
-
+recursive-exclude **/ *.pyc


newville:
@smartsammler Oh, thanks, that's a good catch.  I'll merge this now.  
I guess this warrants a release of 3.5.2, which I'll try to do this week.

smartsammler:
@newville That would be great. Thank you very much.

PR #248, opened 2023-10-17
Prevented unsubscribing events from a cleared channel

### LINKED ISSUE #247: Application using pyepics may crash after epics.ca.clear_cache()
This issue requires use of `epics.ca*` with auto-monitoring enabled and `epics.ca.clear_cache()` at the same time with a little bit of timing "luck". If conditions are right, the application may crash after using `epics.ca.clear_cache()` if it used `epics.ca*()` (e.g. `epics.caget()`) beforehand.

GDB stack trace:

image

hanak:
Closes #247 

This update prevents a random crash (SIGSEGV) after using ca.clear_cache. This crash is caused by calling ca_clear_subscription (done in pv.PV.disconnect when a monitor is used) any time after ca_clear_channel (done in ca.clear_cache) on the same channel. The randomness is caused by CG because pv.PV.disconnect is called from pv.PV.__del__. This update resolves the problem by disconnecting all PVs found in the PV cache prior clearing all channels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,6 @@ dist
 NonGit
 venv
 .vscode
+.coverage
 pyepics.egg-info
 epics.egg-info
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -493,16 +493,20 @@ def clear_cache():
     detaches from the CA context.  This is important when doing
     multiprocessing (and is done internally by CAProcess),
     but can be useful to fully reset a Channel Access session.
+
+    Any class `PV` created prior this call without using `get_pv()`
+    have to be disconnected (use `PV.disconnect()`) explicitly
+    because disconnection clears subscriptions to events of
+    Epics CA connection which are to be cleared here.
+    No instance of the class `PV` created prior this call should be
+    used after because underlaying Epics CA connections are cleared here.
+    Failing to follow these rules may cause your application to experience
+    a random SIGSEGV from inside Epics binaries.
     """
     # Clear the cache of PVs used by epics.caget()-like functions
-    # by disconnecting PVs so that all subscriptions are released
-    # prior clearing channels.
+    # before clear channels references from the cache.
     from . import pv
-    pv_cache = pv._PVcache_
-    pv._PVcache_ = {}
-    for pv in pv_cache.values():
-        pv.disconnect()
-    pv_cache.clear()
+    pv.clear_pvcache()
 
     # Unregister callbacks (if any)
     for chid, entry in list(_chid_cache.items()):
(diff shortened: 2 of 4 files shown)


newville:
@hanak Thanks very much!  In looking at this, I might suggest that we add a `clear_pvcache()` function to `pv.py` and call that from `ca.py`.   Does that seem reasonable to you?    
Otherwise, I would vote that we just merge this.

hanak:
I think `clear_pvcache()` sounds good. Let's add it.

Additional throughts: There is an import loop (`pv` imports `ca` and `ca` imports `pv` in `clear_cache()`) which will stay as is. I think a solution of the loop would be to remove `ca.clear_cache()`, add `ca.clear_cacache()`, and call `ca.clear_cacache()` from `pv.clear_pvcache()`. Yet, this is a breaking change for those who use `ca.clear_cache()`. Still, it would make the structure better going forward. Would you allow such a change (in future)?

hanak:
@newville I am sorry for late response. I agree with you regarding importing `_PVcache_` directly to `ca.py` and manipulating it there is sort of ugly. So, the update you have suggested earlier (`pv.clear_pvcache()`) makes perfect sense to me because this will keep the knowledge at its source (`pv.py`).

Moving `_PVcache_` to `ca.py` will remove the circular import. However, it would inject knowledge about something from a higher layer (`pv.py`) to the lower layer (`ca.py`). At the end, `ca.py` does not care about how the created Epics connections are utilized. It just creates them and maintains them. On the other hand, `pv.py` is one of the users of Epics connection and handles its own cache (`_PVcache_`). So, keeping `_PVcache_` in `pv.py` make sense. Yet, `pv.py` might not be the only consumer of `ca.py` services (in theory). So, what about following:

Allow `ca.py` to clear caches of anything which utilized `ca.py` without explicit knowledge. In case of `pv.py` this would mean: `ca.register_clear_cache(clear_pvcache)`. This way, `ca.py` does not need to have specific knowledge about `pv.py` or any other alternatives to `pv.py`. And, no change to API is required: `ca.clear_cache()` would work as it does today. And, no circular import would be necessary. And, if some other consumer of `ca.py` services is interested, it can join the flow.

At this very moment, it feels clean to me. Yet, I might be missing something. What do you think?

newville:
@hanak Yes, that is a great idea:  Add a "register cache" function in `ca.py` so that other modules, say  `pv.py`, can call to register a cache that `ca.py` can then clear.  Yes, please!

hanak:
@newville I have added the registration function and removed the circular import. Please, review.

newville:
@hanak Thanks!  That looks great to me.  I would suggest we merge this as it is.  Do you agree this is ready?   I'll plan on waiting a day or two just to see if anyone else has comments.

PR #250, opened 2023-10-18
Added pull-request to the list of events which trigger a workflow

### LINKED ISSUE #249: GitHub worflow tests do no execute on pull-requests
Problem: Workflow is set to run on "push". However, this happens only on the branch one is pushing to. In case of forks, it is a branch at the fork. However, results of such a run are not visible from the pull request. This could postpone resolving a test failure to a next fix in line.

hanak:
Closes #249

Enable workflow execution on pull request (open, synchronization, reopen). This makes results of the test visible even for pull-requests from a forked repository. Without this, the results are visible only at the forked repository.

### DIFF
--- a/.github/workflows/test-with-conda.yml
+++ b/.github/workflows/test-with-conda.yml
@@ -1,6 +1,8 @@
 name: Test with Conda, Py3.7 to 3.11
 
-on: [push]
+on:
+  push:
+  pull_request:
 
 jobs:
   build-linux:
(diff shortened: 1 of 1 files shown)


hanak:
Unfortunately, according to the documentation (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows), the events apply only if the configuration is in the default (master) branch. I will try it on the fork first to see whether this works or now.

newville:
@hanak yes, we should definitely automate the testing of PRs. Thanks!

PR #254, opened 2024-01-13
Device/SRS570: fix attrs bug

guidoiaquinti:
## Description
We are missing a comma in the `attrs` tuple resulting into an implicit string concatenation.


newville:
@guidoiaquinti Thanks!

PR #255, opened 2024-01-14
Fix text encoding

### LINKED ISSUE #242: text encoding fallback
It is the same issue as described in #87: epics templates contain latin1(iso-8859-1) characters in the DESC/EGU fields. In our cases they could be "°C" for temperature or "µA" for current. 

The current implemented solution is to let user define *PYTHONIOENCODING* as an environment variable. It has two issues,
* This environment variable has the side effect of changing the default encoding for all stdio and file operations in Python[1].
* A site could have mixed encodings of 'utf-8' and a local specific encoding, e.g. 'latin1' for west europe, 'gb18030' for Chinese, etc. Forcing either one would upset the other.

One way as I see is to use 'utf-8' by default but have a configurable fallback encoding. I implemented this scheme in a test branch[2].
* (+) It is biased towards west European user: the default fallback is "latin1". So no behavior change for the original issue reporter.
* (+) ca get handles bytes with mixed encodings.
* (-) ca put converts strings only to 'utf8' bytes. In case other encoding is needed, the user has to provide the encoded bytes, e.g.
   ```
   ## utf-8 by default
   >>> epics.caput('ao.EGU', '°C')
   1
   >>> epics.caget('ao.EGU$')
   array([194, 176,  67,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
   >>> epics.caget('ao.EGU')
   '°C'
   ## latin1 by choice
   >>> epics.caput('ao.EGU$', '°C'.encode('latin1'))
   1
   >>> epics.caget('ao.EGU$')
   array([253,  67,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
   >>> epics.caget('ao.EGU')
   '°C'
   ```
In a Chinese environment, the fallback encoding is set to "gb18030".
```
$ export PYEPICS_FALLBACK_ENCODING=gb18030
$ python
>>> import epics
>>> epics.caput('ao.EGU', '摄氏度')
1
>>> epics.caget('ao.EGU')
'摄氏度'
>>> epics.caput('ao.EGU', '摄氏度'.encode('gb18030'))
1
>>> epics.caget('ao.EGU')
'摄氏度'
>>> epics.caget('ao.EGU$')
array([201, 227, 202, 207, 182, 200,   0,   0,   0,   0,   0,   0,   0,
```

How do you think of this approach?

[1] https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING
[2] https://github.com/xiaoqiangwang/pyepics/tree/str_fallback_encoding

guidoiaquinti:
## Description
This is an initial attempt to fix #242. A few points to note:

1. I'm new to this repository (hello everyone! 👋). My familiarity with both code and its underlying concepts (including epics itself), is still developing.
2. I didn't find a way to successfully run the tests locally
3. While this commit does not resolve several other file handling bugs (for instance, files are not explicitly closed in some cases), I plan to address these in separate PRs.

### DIFF
--- a/epics/_version.py
+++ b/epics/_version.py
@@ -15,6 +15,7 @@ import os
 import re
 import subprocess
 import sys
+from epics.utils import IOENCODING
 
 
 def get_keywords():
@@ -138,7 +139,7 @@ def git_get_keywords(versionfile_abs):
     # _version.py.
     keywords = {}
     try:
-        f = open(versionfile_abs, "r")
+        f = open(versionfile_abs, "r", encoding=IOENCODING)
         for line in f.readlines():
             if line.strip().startswith("git_refnames ="):
                 mo = re.search(r'=\s*"(.*)"', line)
--- a/epics/autosave/save_restore.py
+++ b/epics/autosave/save_restore.py
@@ -30,6 +30,8 @@ import os
 import datetime
 import json
 from epics.pv import get_pv
+from epics.utils import IOENCODING
+
 
 def restore_pvs(filepath, debug=False):
     """
@@ -42,7 +44,7 @@ def restore_pvs(filepath, debug=False):
     pv_vals = []
     failures = []
     # preload PV names and values, hoping PV connections happen in background
-    with open(filepath, 'r') as fh:
+    with open(filepath, 'r', encoding=IOENCODING) as fh:
         for line in fh.readlines():
             if len(line) < 2 or  line.startswith('<END') or line.startswith('#'):
                 continue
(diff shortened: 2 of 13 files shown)


newville:
@guidoiaquinti Thanks.  To run the test, one does need a working EPICS IOC.  See https://github.com/pyepics/pyepics/blob/master/tests/README
and/or 
https://github.com/pyepics/pyepics/blob/master/.github/workflows/test-with-conda.yml 
to see how these are run.

It's not 100% clear to me that adding an explicit `encoding` to every `open()` is actually necessary, but it seems like it will rarely cause problems where it is unneeded.

PR #261, opened 2024-06-06
Alarm.__repr__ references pv.name instead of pv.pvname

NichtJens:
## Description
Currently printing an `Alarm` object does not work since its `__repr__` references `pv.name` instead of `pv.pvname`.

### DIFF
--- a/epics/alarm.py
+++ b/epics/alarm.py
@@ -117,7 +117,7 @@ class Alarm(object):
         self.check_alarm()
 
     def __repr__(self):
-        return "<Alarm '%s', comp=%s, trip_point=%s >" % (self.pv.name,
+        return "<Alarm '%s', comp=%s, trip_point=%s >" % (self.pv.pvname,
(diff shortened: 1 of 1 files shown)


newville:
I am not sure I understand the error, but merging.  We should drop support for 3.7 anyway.  And, I hope to tag a new release  next week.

NichtJens:
Ah, maybe worth noting: I think the error happened in the 3.10 tests (i.e., not 3.7).

image

But it looks a bit odd:

```python
assert 'R-hc`]OZTjcb...f7ka`<@L9Te<B' == ''
```

I suppose the dots come from shortening for the print out(?)

But the test clearly expects an empty string instead of the random(?) characters:
```python
assert zerostr.get(as_string=True) == ''
```

In any case, it doesn't look like my change could have caused it :)
But it's definitely odd that it works as intended for 3.9 and 3.11 ...

PR #267, opened 2024-06-17
(#266) On Py3.8, install and import importlib_resources instead of importlib.resources

### LINKED ISSUE #266: importlib.resources.files not available on py3.8
In a downstream library I am getting errors on python 3.8 because `AttributeError: module 'importlib.resources' has no attribute 'files'`, which I believe is due to this change: https://github.com/pyepics/pyepics/commit/919ff8dbbc145ba45232ffabc1ff7e73e33d541d

I think the canonical solution is to use https://pypi.org/project/importlib-resources/ for pythons older than 3.9

dperl-dls:
## Description
- Adds a dependency on `importlib_resources` if python version is 3.8
- Imports `files` from there if it is not available from `importlib.resources`
Fixes #266

### DIFF
--- a/epics/ca.py
+++ b/epics/ca.py
@@ -27,7 +27,11 @@ import warnings
 from copy import deepcopy
 from collections import defaultdict
 from math import log10
-from importlib import resources
+
+try: # importlib.resources does not yet have files on python 3.8
+    from importlib.resources import files as importlib_resources_files
+except ImportError:
+    from importlib_resources import files as importlib_resources_files
 
 HAS_NUMPY = False
 try:
@@ -280,7 +284,7 @@ def find_lib(inp_lib_name='ca'):
         return dllpath
 
     # Test 2: look in installed python location for dll
-    dllpath = resources.files('epics.clibs') / clib_search_path(inp_lib_name)
+    dllpath = importlib_resources_files('epics.clibs') / clib_search_path(inp_lib_name)
 
     if (os.path.exists(dllpath) and os.path.isfile(dllpath)):
         return dllpath
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,6 +70,7 @@ install_requires =
     setuptools
     numpy
     pyparsing
+    importlib_resources; python_version<="3.8"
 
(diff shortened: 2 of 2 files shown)


newville:
@dperl-dls I think that maybe `importlib_resources` needs to be added as a dependency for Python 3.8 in `setup.cfg` too, probably to have
```
install_requires =
    setuptools
    numpy
    pyparsing
    importlib_resources; python_version<="3.8"
```

Once we merge #268 (which now correctly illustrates the problem), I'll let you or others decide whether to drop Python 3.8 now or rebase (and probably update `setup.cfg`) this to pass all the tests.  Is that acceptable to you?

dperl-dls:
Thanks for fixing the tests. Looks like they pass on 3.8 on this branch after rebasing: https://github.com/dperl-dls/pyepics/actions/workflows/test-python38.yml

I guess since the kludge is small and only has to be there for a few months, we might as well keep support for that time, I think there probably are still people depending on 3.8. 

Thanks a lot for your quick responses on this!

newville:
@dperl-dls Thanks.  Yes, I think we should release this to support the latest changes with Python 3.8.  I don't know how much it is used, but apparently not zero. And pyepics probably should be infrastructure,  so not driving version upgrades.

From my POV, this was helpful in pointing out that specifying the Python version was broken for the CI testing and helped inspire that fix.

tboegi:
I think it used, I got this problem on gha today:

```
AttributeError: module 'importlib.resources' has no attribute 'files'
  File "/usr/local/lib/python3.8/dist-packages/epics/ca.py", line 377, in initialize_libca
    libca = load_dll(find_libca())
  File "/usr/local/lib/python3.8/dist-packages/epics/ca.py", line 333, in find_libca
    return str(find_lib('ca'))
  File "/usr/local/lib/python3.8/dist-packages/epics/ca.py", line 283, in find_lib
    dllpath = resources.files('epics.clibs') / clib_search_path(inp_lib_name)
```
Thanks for working on this.

newville:
@dperl-dls Thanks again!

PR #270, opened 2024-08-20
Str waveforms 1 elem

### LINKED ISSUE #269: camonitor of a string array PV raises a TypeError exception if the array contains one string
Exception:
```
File "...\site-packages\epics\__init__.py", line 157, in camonitor
  thispv.get()
File "...\site-packages\epics\pv.py", line 486, in get
  data = self.get_with_metadata(count=count, as_string=as_string,
File "...\site-packages\epics\pv.py", line 48, in wrapped
  return func(self, *args, **kwargs)
File "...\site-packages\epics\pv.py", line 621, in get_with_metadata
  if count < len(val):
TypeError: len() of unsized object
```

The behavior is a result of interaction between following functions:
* `ca.unwap()` uses `ca.scan_string` to read a string. This function will return the string if retrieved list of strings contains just one string.
* `PV.get_with_metadata(as_numpy=True)` (the default used by `camonitor()`) does following:
  * Reads the string into a variable `val`
  * It tests that `val` supports `__len__`. In case of a string, this passes.
  * Then, it verifies `as_numpy` and `ca.HAS_NUMPY`. In this is satisfied. So, it converts `val` to a numpy object.
  * Then, it tries to run `len(val)`. Yet, `val` is a scalar at this moment (`val.shape` returns `()`). This raises an exception.

Further details:
* This has been observed on pyepics 3.4.3 using for automated tests. Yet, the code involved in this mechanism is the same at the current master branch.
* Standalone recreation of the scenario has not been performed because the PV is a custom record which mimics 
waveform (FTVL=STRING): it provides an array of strings. caget works with no complains.
* Use of `as_numpy=False` with `caget()` will return the first letter of the first string (i.e. `"R"` instead of `"READY"`) for an array containing `["READY"]`.

newville:
## Description

Fixes #269

### DIFF
--- a/tests/test_pv_unittests.py
+++ b/tests/test_pv_unittests.py
@@ -306,16 +306,12 @@ def test_string_waveform_lengths():
     readback = spv.get()
     nelm = spv.nelm
     with no_simulator_updates():
-        n = 0
-        for count in (nelm, nelm-2, nelm//3, 11, 8, 5, 3, 2, 1):
-            n += 5*count
-            dat = [f"str_{i+n}" for i in range(count)]
+        for count in (nelm, nelm-5, nelm//3, 11, 5, 1, 3, 7, 2, 1):
+            dat = [f"str_{count}_{i}" for i in range(count)]
             spv.put(dat)
-
             time.sleep(0.1)
             readback = spv.get()
             assert all(readback == dat)
-            #            print(count, dat, readback)
 
 
 def test_subarrays():
--- a/tests/test_pv_unittests.py
+++ b/tests/test_pv_unittests.py
@@ -6,7 +6,7 @@ import time
 import numpy
 import threading
 import pytest
-
+from random import random
 from contextlib import contextmanager
 from epics import PV, get_pv, caput, caget, caget_many, caput_many, ca
 
@@ -267,7 +267,7 @@ def test_put_string_waveform_single_element():
         pv.put(put_value, wait=True)
         time.sleep(0.05)
         get_value = pv.get(use_monitor=False, as_numpy=False)
-        assert put_value[0] == get_value
+        assert put_value == get_value
 
 def test_put_string_waveform_mixed_types():
     write('String Array: put mixed types\n')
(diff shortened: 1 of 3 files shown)


newville:
with all tests passing, I will merge within 3 days unless there are objections.

tests/test_pv_unittests.py
```diff
@@ -299,6 +299,25 @@ def test_put_string_waveform_zero_length_strings():
...
+def test_string_waveform_lengths():
+    "test string arrays with length=1 and other lengths"
+    spv = get_pv(pvnames.string_arr_pv)
+    readback = spv.get()
+    nelm = spv.nelm
+    with no_simulator_updates():
+        n = 0
+        for count in (nelm, nelm-2, nelm//3, 11, 8, 5, 3, 2, 1):
+            n += 5*count
+            dat = [f"str_{i+n}" for i in range(count)]
```
hanak, line 312: Can we just use the string `f"str_{count}_{i}"`? It seems to me this will remove the need to the variable "n" and it might help to guess which loop put the values in just by looking at the value.

tests/test_pv_unittests.py
```diff
@@ -299,6 +299,25 @@ def test_put_string_waveform_zero_length_strings():
         get_value = pv.get(use_monitor=False, as_numpy=False)
         numpy.testing.assert_array_equal(get_value, put_value)
 
+
+def test_string_waveform_lengths():
+    "test string arrays with length=1 and other lengths"
```
hanak, line 304: Shouldn't we use `"""`?

tests/test_pv_unittests.py
```diff
@@ -299,6 +299,25 @@ def test_put_string_waveform_zero_length_strings():
...
+        n = 0
+        for count in (nelm, nelm-2, nelm//3, 11, 8, 5, 3, 2, 1):
+            n += 5*count
+            dat = [f"str_{i+n}" for i in range(count)]
+            spv.put(dat)
+
+            time.sleep(0.1)
+            readback = spv.get()
+            assert all(readback == dat)
+            #            print(count, dat, readback)
```
hanak, line 318: Assuming pytest is used, the print and the assert can be merged into one using:
assert readback == dat, f"Count {count}"

PR #272, opened 2024-09-27
Fixes autosave issues

jbhopkins:
Fixes issues where autosave:
- Wouldn't restore values if the pv had no value in the save file
- Wouldn't load .req files with macro definitions (e.g. P=)
- Wouldn't propagate macro definitions to a .req file loaded within another file

For the last item, suppose you have first.req which has the line:
file "second.req", P=IOC1

and second.req has the line:
file "third.req", P=$(P)

The value of P (IOC1) from first.req wouldn't propagate into third.req.

### DIFF
--- a/epics/autosave/save_restore.py
+++ b/epics/autosave/save_restore.py
@@ -46,7 +46,8 @@ def restore_pvs(filepath, debug=False):
     # preload PV names and values, hoping PV connections happen in background
     with open(filepath, 'r', encoding=IOENCODING) as fh:
         for line in fh.readlines():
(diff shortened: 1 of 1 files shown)


newville:
@jbhopkins Thanks very much.  I'll merge this, and also add to my ToDo list:

1. add at least a basic test with latest autosave.   We did have some tests (in "UnsortedTests") long ago, but these are not included in the GitHub Actions testing..... we should fix that.
2. maybe we need to check out more elaborate Save/Restore examples  - it's been a long time since I looked at that!

PR #273, opened 2024-10-02
Fixes an issue where background color of wx epics widgets with the PMixin couldn't be set

jbhopkins:
As title. There was a typo in the wx epics widgets PVMixin that prevented the background color from being set.


newville:
@jbhopkins Thanks!  That looks like it has been there a long time!

PR #283, opened 2025-12-19
doc: remove unused sphinxcontrib.video plugin

minijackson:
## Description

This removes the seemingly unused `sphinxcontrib.video` plugin from the documentation, which removes one dependency from my packaging ^^


newville:
@minijackson Thanks.  Yes, we are not currently using that extension here. If we do add that back in, we'll also add it to `pyproject.toml`.

PR #286, opened 2026-05-20
fix: use set_parse_action instead of setParseAction

wakonig:
## Description

`pyparsing` deprecated `setParseAction` in favor of `set_parse_action`. 

It looks to me as if these two methods are the same, see  https://github.com/pyparsing/pyparsing/blob/4f52ba7e9daa4845fd94dd4a1999fc42204207fe/pyparsing/core.py#L2604


newville:
@wakonig Thanks!  I had not seen that deprecation or problem, so thanks for catching this.

